PR #2, opened 2015-09-05
Only MIT? [ci skip]

PallHaraldsson:
The exception is gone.

https://tls.mbed.org/foss-license-exception

Cut rest of the text?

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -21,4 +21,4 @@ The Mbed.jl package is licensed under the MIT "Expat" License:
 > TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 > SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-Pursuant to the [ARM mbed FOSS License Exception](https://tls.mbed.org/foss-license-exception), the distribution of libmbedtls, libmbedcrypto, and libmbedx509 ("mbedtls") that is installed as a consequence of installing this package may be redistributed along with this package. The source code of mbedtls may be downloaded in its entirety from https://tls.mbed.org/source-code. 
+The source code of mbedtls may be downloaded in its entirety from https://tls.mbed.org/source-code. 


PallHaraldsson:
I'm not sure why I'm getting "continuous-integration/appveyor — AppVeyor build failed". I think it's a false alarm (like has happened before, and I'm not sure the [ci skip] I just added helps..)

malmaud:
The AppVeyor problem was preexisting (and isn't a false alarm). Thanks for this PR though.

tkelman:
I think you need to re-enable appveyor after moving the repository ownership.

PallHaraldsson:
If you're still reading.. When I think about it, maybe you should mention GPL also in the license..

Mbed TLS is also available under GPL (v.2.0).

In case anyone would think this Apache code dependency would not be compatible with code they use, I think this dual-license thing might be a worthy addition:

Note, Apache considers Apache 2.0 to be compatible with the GPL (all versions), but FSF considers it only compatible with GPLv3 (but still free software).

Since Julia has GPL components (that could be stripped away) and even if you (this library or others using Julia) do not use the GPL libraries, it might be a gray area. Still, you could be willing to use GPL code (actively adding) and then, for sure, your whole combination can't be distributed (except under GPL).

malmaud:
Thanks @tkelman .

@PallHaraldsson AFAIK it's not dual licensed; the license has just switched starting in 2.1.0 to Apache 2.0. Perhaps we can ask for a dual-licensed version, though.

PallHaraldsson:
On mán  7.sep 2015 10:46, Jon Malmaud wrote:
> Thanks @tkelman <https://github.com/tkelman> .
>
> @PallHaraldsson <https://github.com/PallHaraldsson> AFAIK it's not dual
> licensed; the license has just switched starting in 2.1.0 to Apache 2.0.
> Perhaps we can ask for a dual-licensed version, though.

I think you are just wrong on that (and you do not have to ask..). If 
you look further than the subject title:

https://tls.mbed.org/tech-updates/releases/mbedtls-2.1.0-released-under-apache

"Apache licensed
Get your copy here: mbedtls-2.1.0-apache.tgz
[..]
GPL licensed
Get your copy here: mbedtls-2.1.0-gpl.tgz"

They may not use the term "dual licensed", but isn't this effectively 
what it is? I think you are confusing this with the previous linking 
exception where they allowed all free software to use the GPL package. 
That is now not needed for permissive licensed software, but the GPL 
option seems now important in reverse to allow copyleft software to 
(still) use the (otherwise) relicensed software.

Will they still allow GPL after 2.1.0, I don't know, but I guess as the 
same problem would persist otherwise.


See also:
http://www.gnu.org/licenses/license-list.en.html#apache2

And the situation from the Apache side is not what I thought (for GPLv3, 
except I see they do not want it):
http://www.apache.org/licenses/GPL-compatibility.html

malmaud:
Oh great, so it is dual-licensed.

PR #4, opened 2015-09-07
GPL

PallHaraldsson:
I couldn't figure out how to add another commit in my former issue (that is now merged, and maybe then too late?).. Sending as separate pull request. Seems ok, not to combine with the other (or combine if possible).

Note, I didn't say GPLv2 as I'm not sure that is what GPL means there. I guess this vagueness, is ok, people will at least then know and can look up (the file, I was to lazy too). If if is GPLv2 then, be default it can be upgraded to GPLv3 (unless it would say "only").

Only if it says GPLv3, there could be a problem (as not backwards compatible with GPLv2) but I doubt they do that, and anyway, it would strictly not be *your* problem.. You would at least be giving heads-up.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -22,7 +22,7 @@ The Mbed.jl package is licensed under the MIT "Expat" License:
 > SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 The MbedTLS library, which is installed as a consequence of installing this package,
-is licensed separately, as follows:
+is licensed separately, as follows (but also under the GPL):
 
 > Copyright (C) 2006-2015, ARM Limited, All Rights Reserved
 > SPDX-License-Identifier: Apache-2.0


malmaud:
Thanks for the PR. Before merging this, I'd like to get clarification from the MbedTLS developers on the exact nature of the dual license situation since the compiled libraries from the GPL-distributed source and Apache-distributed source are not 100% bit-wise identical. 

It's possible we might have to allow the build process to be passed an option through an environment variable on whether the GPL version of Apache version of the MbedTLS libraries should be built.

malmaud:
I'm seeking clarification on the dual-license situation: https://github.com/ARMmbed/mbedtls/issues/282.

LICENSE.md
```diff
@@ -22,7 +22,7 @@ The Mbed.jl package is licensed under the MIT "Expat" License:
 > SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 The MbedTLS library, which is installed as a consequence of installing this package,
-is licensed separately, as follows:
+is licensed separately, as follows (but also under the GPL):
```
hayd: This is confusingly worded, I think it should be clearer (once we have clarification). At the moment this clarifies nothing and I'm not sure it's even correct (MbedTLS.jl is not using MbedTLS under the GPL ...?).

Potentially also worth adding a link, rather than being brief/ambiguously worded.

malmaud: Ya, this won't be the final wording. If we get permission to dual-license, I'd rather find some standard dual-licensing wording that others have used.

malmaud:
So the answer seems to be that it is permissible to dual-license and use the same binaries. For building via source, it probably is safest to either download the source with the GPL license text or the source with the Apache license text depending on a build option, I guess?

hayd:
Forgive my silly question but what's the benefit of dual licensing here? (and not just using the Apache version which IIUC is more forgiving/compatible with MIT.)

malmaud:
It's so that we could distribute a bundle that has both Julia (which has GPL v2 code by default) with the web stack.

hayd:
> Julia (which has GPL v2 code by default)

Ah, thanks. I didn't realise this was still the case.

PallHaraldsson:
First not there is 2.1.1 out (and I only checked GPL situation for 2.1.0) and:
https://github.com/ARMmbed/mbedtls/blob/development/ChangeLog
"Primary open source license changed to Apache 2.0 license." - implies there is a secondary..

https://tls.mbed.org/tech-updates/releases/mbedtls-2.1.0-released-under-apache
"As announced above, our main open source license is now the Apache 2.0 license. This is also reflected in our source code in the Github source code repository.

We are aware that some project [sic] cannot use an Apache licensed version and used the GPL licensed version before.

For those projects, we provide a GPL licensed package on the download page."

You may want to let them know there is more than one GPL project.. :)

https://tls.mbed.org/contact

I looked into the GPL and Apache distribution, to see if the GPL distribution only differed in the master license file. I see at least one program file is different, but only because the headers are different (GPL vs Apache). I assume but *didn't* confirm that the actual source code and everything else is the same. It would seem strange not to since they share the "same" version number. Still, it would seem a lot of work to maintain this and distribute two versions and having to change license from Apache to GPL in each header. Since they own the source code they could do that forever (they want you to share copyright with them if you contribute), but would they stop doing that?

It would seem wise for them, and someone to point out to them, to have a master file that says "you are allowed to link [dynamically] with GPL code"..

If they break GPL compatibility, it would seem projects they list will be in trouble, and I hope this doesn't apply..:

https://tls.mbed.org/kb/licensing/how-is-mbedtls-protected
"Won't people use mbed TLS or PolarSSL illegally?
Yes, they probably will.
[..]
If we ever find out, we will of course prosecute them to the fullest extent of the law."


That and some of the stuff on their page is clearly (at least partially) outdated - from GPL and FOSS Exception era:

https://tls.mbed.org/kb/generic/projects-using-mbedtls
E.g. PowerDNS is listed:


https://github.com/PowerDNS/pdns
"PowerDNS is copyright Ⓒ 2002-2015 by PowerDNS.COM BV and lots of contributors, using the GNU GPLv2 license (see NOTICE for the exact license and exception used)." [about OpenSSL]

Still I see in their commit from August:

goodby PolarSSL, hello mbed TLS
https://github.com/PowerDNS/pdns/commit/8cb5dea2ce601a8e79f606f7af9e7676b1feee40

Update mbed TLS to version 2.1.0 (GPL)
https://github.com/PowerDNS/pdns/commit/437d6216e21b21501ba19e579aed71b168dbff41
[ext/mbedtls/LICENSE]
"+This package of mbed TLS is specifically licensed under the GPL 2.0,
+as can be found in: gpl-2.0.txt"


Something to keep in mind later..:
https://tls.mbed.org/kb/development/thread-safety-and-multi-threading
"mbed TLS is designed to be used in threaded and non-threaded environments. In order to keep mbed TLS thread-safe it is important to keep a few things in mind."

malmaud:
I added text to the license in https://github.com/JuliaWeb/MbedTLS.jl/commit/aec195dbcbad4520dd18d5d2948e0e1c699c46ff to clarify the dual-licensing situation.

PR #8, opened 2015-10-21
build only library

wildart:
and do not build examples and tests


malmaud:
LGTM; test failures are upstream Homebrew issues.

PR #20, opened 2016-01-21
Explicitly check for cmake

malmaud:


yuyichao:
Does this mean that the build process will raise an error if cmake is not found but the library is installed system-wise?

malmaud:
Mmm ya, I guess you're right about that.

tkelman:
add a shell script to do the checking with a nice error and have it run as the first build step?

malmaud:
@tkelman how's this?

deps/cmake_check.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+command -v cmake > /dev/null
+
+if [[ $? -ne 0 ]]; then
```
yuyichao: FYI, this can be simplified to `if command -v cmake > /dev/null; then`.

yuyichao:
LGTM.

malmaud:
Thanks @yuyichao

tkelman:
this shouldn't assume bash.

PR #21, opened 2016-01-26
Fix handshake test due to overly type-restricted Base method

jrevels:
This PR allows tests to pass when running against the current Julia master branch.

The error was because the [`sz` parameter passed to `readbytes!` here](https://github.com/JuliaWeb/MbedTLS.jl/blob/master/src/ssl.jl#L99) is a `UInt64`. This value eventually [gets passed to `wait_nb`](https://github.com/JuliaLang/julia/blob/master/base/stream.jl#L921), but that function's argument is [restricted to `Int`](https://github.com/JuliaLang/julia/blob/master/base/stream.jl#L358).

While that restriction should probably be loosened to `Integer` (plus a positivity check), it's easy to patch over it here by converting `sz` to `Int` before passing it to `readbytes!`.

@kshyatt

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -99,7 +99,7 @@ end
 function f_recv(c_ctx, c_msg, sz)
     jl_ctx = unsafe_pointer_to_objref(c_ctx)
     jl_msg = pointer_to_array(c_msg, sz, false)
-    n = readbytes!(jl_ctx, jl_msg, sz)
+    n = readbytes!(jl_ctx, jl_msg, Int(sz))
     Cint(n)
 end
 


kshyatt:
Thank you very much for this. Now we can finally run coverage again.

tkelman:
Not actually necessary any more, ref #19.

jrevels:
Damn, second time today I got tony-ninja'd. Sorry for the noise.

tkelman:
Test cases for base that are representative of what this package needs would be awesome for hopefully making future breakage a little less likely.

PR #23, opened 2016-01-29
don't assume bash

tkelman:
for various bsd's or unusual distros that may not have bash or have it in an unusual location

### DIFF
--- a/deps/cmake_check.sh
+++ b/deps/cmake_check.sh
@@ -1,8 +1,6 @@
-#!/bin/bash
+#!/bin/sh
 
-command -v cmake > /dev/null
-
-if [[ $? -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


deps/cmake_check.sh
```diff
@@ -1,8 +1,6 @@
-#!/bin/bash
+#!/bin/sh
 
-command -v cmake > /dev/null
-
-if [[ $? -ne 0 ]]; then
+if cmake --version > /dev/null; then
```
yuyichao: `if cmake --version > /dev/null 2>&1; then` since there's separated error printing below.

tkelman: `if !` actually, but sure

yuyichao: Ahh, right. LGTM.

PR #24, opened 2016-01-30
Adjustments for IO refactoring

malmaud:


REQUIRE
```diff
@@ -1,3 +1,3 @@
-julia 0.4-
+julia 0.5-
```
tkelman: are we going to maintain separate branches or is this just for testing?

malmaud:
I was thinking separate branches.

tkelman:
That could get messy. If we can manage to get all of this via compat that might be less ongoing maintenance burden.

malmaud:
Sure. I probably won't look into that myself for some time, though.

Keno:
Can we get this merged? Requests seems broken for HTTPS connections without this.

malmaud:
I’ll work on getting the necessary changes to Compat merged today so that we can merge this without needing separate branches.

PR #25, opened 2016-01-31
More adjustments for IO refactoring

ihnorton:
...


malmaud:
Thanks!

malmaud:
Oh actually, we should decide if we're maintaining .4 compatibility on master before merging this.

tkelman:
I'm reasonably sure this package is going to need 0.4-compatible bugfix releases. Unless fixing it on master is really urgent, I think getting https://github.com/JuliaLang/Compat.jl/pull/162 into shape and seeing if that can be used here is a better option.

malmaud:
Makes sense

PR #27, opened 2016-02-03
Remove duplicate definitions

MichaelHatherly:
The `CipherInfo` duplicates had slightly different error messages. Other than that they were identical.

The signatures for the `parse_key!` methods were written such that

    parse_key!(ctx::PKContext, key, pw) = parse_key!(ctx, key, Nullable(pw))

calls itself resulting in a stack overflow eventually. Restricting `maybe_pw` in the first method to `Nullable` fixes that.

### DIFF
--- a/src/cipher.jl
+++ b/src/cipher.jl
@@ -163,13 +163,6 @@ function CipherInfo(id::CipherID)
     end
 end
 
-function CipherInfo(kind::CipherKind)
-    ptr = ccall((:mbedtls_cipher_info_from_type, MBED_CRYPTO), Ptr{Void},
-        (Cint,), Int(kind))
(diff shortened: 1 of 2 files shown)


tkelman:
Whoa, so does deleting unused code magically fix the package on nightly? Without needing to break it on 0.4 like #24 would?

MichaelHatherly:
Yeah, not sure how it's managed to fix things, if that's actually the case.

malmaud:
Thanks @MichaelHatherly, those mistakes were artifacts of the refactoring I did earlier.

tkelman:
Oh right, I guess it isn't this package that's broken, but Requests.

PR #30, opened 2016-02-11
add compatibility to unsafe_read/write additions to base julia in v0.5

vtjnash:

PR #37, opened 2016-04-23
be more specific with unsafe_read/unsafe_write version cutoff

tkelman:
to avoid breaking the package on older 0.5-dev versions, ref #30


malmaud:
Thanks, glad we got this sorted out.

PR #44, opened 2016-05-28
String compat fixes

malmaud:


deps/build.jl
```diff
@@ -7,7 +7,7 @@ function validate_mbed(name, handle)
         get_version = Libdl.dlsym(handle, :mbedtls_version_get_string)
         version_ptr = Vector{UInt8}(9)
         ccall(get_version, Void, (Ptr{Void},), version_ptr)
-        version = VersionNumber(bytestring(pointer(version_ptr)))
+        version = VersionNumber(String(pointer(version_ptr)))
```
malmaud: This isn't correct

malmaud: Actually it is.

tkelman:
Probably needs minimum Compat version to be bumped to more recent than 0.7.9?

malmaud:
I think as it happens this will work, but I should make another PR soon
that bumps it and switches to using unsafe_string.

PR #57, opened 2016-08-01
Add ALPN support

sorpaas:
This pull request add ALPN support for the wrapper. According to HTTP/2 spec, this is necessary for negotiating protocols when HTTPS is enabled.

For the test, I used "google.com" because "httpbin.org" doesn't support HTTP/2 for now.


malmaud:
Thanks!

PR #58, opened 2016-08-02
Fix build dep warns

ranjanan:
Some of the depwarns are also coming from Homebrew.jl, and I'm fixing them on a separate PR (https://github.com/JuliaLang/Homebrew.jl/pull/140)


malmaud:
Thanks!

ranjanan:
@malmaud after that Homebrew PR gets merged, and a new version is tagged for that, could you tag a new version that uses that? (I'll ping again after, I suppose). This would make the `Pkg.add` experience a lot more smooth without all the annoying depwarns.

ranjanan:
@malmaud Can you tag a new version of this package in METADATA?

PR #59, opened 2016-08-03
Add appropriate lower bounds for BinDeps and Homebrew

ranjanan:
[ci skip]

Adding the right lower bounds for BinDeps and Homebrew. Should fix most if not all of the dep warns during install. 

cc: @tkelman

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.4
-BinDeps
+BinDeps 0.4.2
 Compat 0.8.0
-@osx Homebrew
+@osx Homebrew 0.3.3


tkelman:
If it's only fixing warnings, then lower bounds aren't needed. Only if the package actually breaks on lower versions than the bound is this strictly required.

tkelman:
So if you can find the lowest versions of BinDeps below which the package wouldn't install/work at all, those are the right versions for the lower bound.

ranjanan:
Oh alright, you're right. I guess this PR can be closed then.

tkelman:
The package manager in normal operation will try to find the highest allowed versions when you do update, add, resolve, etc. So allowing versions that may have warnings would allow for more feasible solutions in case some packages have upper bounds for whatever reasons. In most cases, if there aren't upper bounds, you'll get the newer warning-free versions like you want.

PR #60, opened 2016-08-03
Add an AptGet provider in order to use the system's MbedTLS

### LINKED ISSUE #55: Avoid download/build if system library is already installed
MbedTLS doesn't build out of the box on my system (debian unstable), but putting that aside I already have a the system library installed and development headers available, so it should *really* try to use the system package as opposed to build a new version of the library.

Is there a way to do that?

wavexx:
Fixes #55

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -50,6 +50,13 @@ if is_unix()
                 end
             end
         end), mbed_all, installed_libpath=joinpath(mbed_dir, "library"))
+
+    if is_linux()
+        provides(AptGet,
+            Dict("libmbedtls10"=>mbed,
+                 "libmbedcrypto0"=>mbed_crypto,
+                 "libmbedx509-0"=>mbed_x509))
+    end
 end
 
 if is_apple() 


malmaud:
Thanks!

tkelman:
Are we sure the debian/ubuntu system copies of mbedtls are always going to be the right version with the same API? Should add a much more specific validation hook if not. Also you're several versions behind the latest release here, since this is a security-critical library it's really important to update frequently.

wavexx:
On Thu, Aug 04 2016, Tony Kelman <notifications@github.com> wrote:
> Are we sure the debian/ubuntu system copies of mbedtls are always
> going to be the right version with the same API? Should add a much
> more specific validation hook if not. Also you're several versions
> behind the latest release here, since this is a security-critical
> library it's really important to update frequently.

How the system's version is managed shouldn't be a concern here, this is
delegated to the sysadmin. The idea is that I can manage the same SO for
all systems, and this is *much* more reliable than checking every single
julia instance.

This also avoid endless build problems. The approach of having the so
built as part of Pkg is only sensible for a single user scenario with a
standard setup. I need to maintain a /cluster/ with a plethora of
programs using MbedTLS.

ALWAYS allow to build external libraries OUT of the Pkg's build system.
I've moved over dozens of built-in package managers for various
languages, and I'm just tired of the silliness of assumptions are baked
into the various package's build process and working around them is is
just tiring. It always works poorly. It's fine for quick development,
but not for production.

The SO is checked with the same rules as the hand-built library. We
could add additional restrictions on the package version, but then
again, I wouldn't rely on the package version at all. What is being done
here (checking the abi version) is already correct way to go.

tkelman:
If you have a system copy already installed that satisfied the requirements, BinDeps should find it and be allowed to use it already. I'm expressing concern that adding an AptGet provider will lead to more people being asked for a root password when installing this package, and trying to install outdated versions of the library from distribution packages that won't work. All the sysadmins in the world won't make debian any less conservative about what they package.

wavexx:
On Thu, Aug 04 2016, Tony Kelman <notifications@github.com> wrote:
> If you have a system copy already installed that satisfied the
> requirements, BinDeps should find it and be allowed to use it already.

It really didn't for me, despite being available. BinDeps kept
downloading and trying to build the package. I need to specify extra
parameters to configure and setup the environment appropriately, so the
default build will always fail.

I _need_ to be able to build out of tree.
What should I look into?

> I'm expressing concern that adding an AptGet provider will lead to
> more people being asked for a root password when installing this
> package, and trying to install outdated versions of the library from
> distribution packages that won't work. All the sysadmins in the world
> won't make debian any less conservative about what they package.

Does BinDeps attempts to actually fetch the dependency? (it doesn't for
me).
This is definitely also something that I'd like to avoid as well.
I really don't need the Apt provider if the library detection works as
intended.

As for the rest, keep in mind both ends. Scripts running using the
locally-built version of Julia will likely never receive updates also
once they've been put somewhere. Pkg updates depend on the end user
being an active developer.

tkelman:
The C code and the Julia bindings really need to move in lockstep almost all of the time. So if you're updating one but not the other, that's not going to work well. If you have your own local deb repository and are creating deb's from Julia packages too, then you can patch whatever you want, or distribute a generated deps.jl file.

BinDeps is annoyingly hard to debug sometimes, but it should be searching global library paths and using anything that passes the validation checks that it finds. I believe system paths are a higher default search priority than building from source. Maybe throw in some `@show` lines during the validation function or inside BinDeps to see if it's checking your system-wide `/usr/` locations. May just need aliases if your global deb packages have sonames and you don't have the dev packages installed?

malmaud:
Let me just revert this for now until the discussion pans out, since @tkelman raises some good points.

wavexx:
On Thu, Aug 04 2016, Tony Kelman <notifications@github.com> wrote:
> The C code and the Julia bindings really need to move in lockstep
> almost all of the time. So if you're updating one but not the other,
> that's not going to work well. If you have your own local deb

I'm aware of this.
I wrote and I'm maintaining other bindings too.

> if it's checking your system-wide `/usr/` locations. May just need
> aliases if your global deb packages have sonames and you don't have
> the dev packages installed?

Everything is installed already, as I'm building other programs with
MbedTLS and using it for internal development as well.

Can you hint at where you would put some @shows to debug the issue?

tkelman:
Definitely show the name and handle in `validate_mbed` here https://github.com/JuliaWeb/MbedTLS.jl/blob/afd2ef93ca24ff231f1c0fb656644f50bf1ac14e/deps/build.jl#L7 - I'm not sure if there's a way to get an absolute path from a handle but that would be worth looking at.

You can try `BinDeps.debug("MbedTLS")` and see what that says as well. Various steps in https://github.com/JuliaLang/BinDeps.jl/blob/68ce26216445bea3a8862ffee2f7f48a08b27c99/src/dependencies.jl#L491-L557 and the other functions in that file would be worth looking at.

wavexx:
I tried on a fresh system, and it seems that the package doesn't build with Julia 0.5:

````
INFO: Installing MbedTLS v0.2.5
INFO: Building MbedTLS
WARNING: Could not check MbedTLS version: MethodError(convert,(String,Ptr{UInt8} @0x00007f710c8754f0))
WARNING: Could not check MbedTLS version: MethodError(convert,(String,Ptr{UInt8} @0x00007f710d3acdd0))
WARNING: Could not check MbedTLS version: MethodError(convert,(String,Ptr{UInt8} @0x00007f710d3ad250))
WARNING: Could not check MbedTLS version: MethodError(convert,(String
````

This has been fixed in the current master though. This results in a double failure (first check on the system libraries fails, then builds the source [optionally fails] and then fails on the second check again.

PR #65, opened 2016-08-10
Also add MD write methods for byte I/O

Keno:
Accidentally left out of my last PR. Will add as soon as CI says it's green.


tkelman:
not covered in tests?

PR #66, opened 2016-08-14
Use system MbedTLS

malmaud:


deps/build.jl
```diff
@@ -31,12 +31,28 @@ else
     srcsha = "8f25b6f156ae5081e91bcc58b02455926d9324035fe5f7028a6bb5bc0139a757"
 end
 
+# Use the version of MbedTLS that ships with Julia .5 and later
+
+julia_bin = Base.julia_cmd().exec[1]
```
tkelman: just use `JULIA_HOME`, that's what this is derived from anyway - the representation of `Cmd` might change which could break this

deps/build.jl
```diff
@@ -31,12 +31,28 @@ else
...
+
+# For a binary install of Julia
+provides(Binaries,
+    joinpath(dirname(julia_bin), "..", "lib", "julia"),
+    mbed_all)
+
+# For a source install of Julia
```
tkelman: don't bother - you should do `make install` before doing anything with packages

deps/build.jl
malmaud: Done

deps/build.jl
malmaud: Done

malmaud:
Travis is showing this error only on OS X. Maybe something wrong internally with BinDeps on Travis for Apple?

```
LoadError: MethodError: no method matching generate_steps(::BinDeps.LibraryDependency, ::BinDeps.CustomPathBinaries, ::Dict{Symbol,Any})
Closest candidates are:
  generate_steps(::BinDeps.LibraryDependency, !Matched::BinDeps.AptGet, ::Any) at /Users/travis/.julia/v0.6/BinDeps/src/dependencies.jl:320
  generate_steps(::BinDeps.LibraryDependency, !Matched::BinDeps.Yum, ::Any) at /Users/travis/.julia/v0.6/BinDeps/src/dependencies.jl:331
  generate_steps(::BinDeps.LibraryDependency, !Matched::BinDeps.Pacman, ::Any) at /Users/travis/.julia/v0.6/BinDeps/src/dependencies.jl:343
  ...
while loading /Users/travis/.julia/v0.6/MbedTLS/deps/build.jl, in expression starting on line 83
```

tkelman:
CustomPathBinaries is probably not widely tested with local paths. I'm not too familiar with where in BinDeps that code is.

malmaud:
Alright. Realistically speaking, I'm probably not going to look into
patching BinDeps personally so this will remain open for the time being.

deps/build.jl
```diff
@@ -31,12 +31,21 @@ else
     srcsha = "8f25b6f156ae5081e91bcc58b02455926d9324035fe5f7028a6bb5bc0139a757"
 end
 
+# Use the version of MbedTLS that ships with Julia .5 and later
+
+provides(Binaries,
+    joinpath(dirname(JULIA_HOME), "..", "lib", "julia"),
```
tkelman: One of either `dirname` or `".."` are redundant, `JULIA_HOME` is already the `bin` directory. On Windows this should be just `JULIA_HOME`.

deps/build.jl
```diff
@@ -31,10 +31,22 @@ else
...
+        BinDeps.ChecksumValidator(get(opts,:SHA,get(opts,:sha,"")),h.path)
+    end
+end
+
 # Use the version of MbedTLS that ships with Julia .5 and later
 
+# For Unix
+provides(Binaries,
+    joinpath(JULIA_HOME, "..", "lib", "julia"),
+    mbed_all)
```
tkelman: `os = :Unix` etc

malmaud:
On Travis, it looks like MbedTLS is still being build from source on Linux and from Homebrew on OS X, even though the final file paths in `deps.jl` are correctly pointing the versions bundled with Julia. So that's a behavior of `BinDeps` that I don't really undestand.

PallHaraldsson:
I do see "mbedtls (>= 2.2)" at Julia's README.md, while 2.2 is no longer listed as the supported version.

I made this https://github.com/JuliaWeb/MbedTLS.jl/issues/67 issue before seeing this one. You think Julia will better track security updates or just want to avoid duplicates? I did wander [for cURL], is that at least possible? Maybe better to track security updates (or not..)? [It's also a possibility that security updates for 0.4 might get neglected..]

malmaud:
The versions of Julia without a bundled MbedTLS are failing on Travis: "LoadError: Provider BinDeps.Binaries failed to satisfy dependency libmbedtls". 
I thought BinDeps would automatically try a different provider in that case? But I guess not and we need to check the Julia version before adding that binary provider?

tkelman:
My impression was that it should also check several providers in a priority order and fall back to other options, but that might be broken?

malmaud:
So that's one question, and the other is why the linux nightly is still building MbedTLS from source, and why Homebrew is still downloading an MbedTLS on the OS X nightly, even though in the latter two cases the final paths in `deps.jl` actually do point to the Julia-bundled binaries. Those all seem like issues with `BinDeps`.

deps/build.jl
```diff
@@ -31,12 +31,34 @@ else
...
+    steps = @build_steps begin
+        BinDeps.ChecksumValidator(get(opts,:SHA,get(opts,:sha,"")),h.path)
+    end
+end
+
+# Use the version of MbedTLS that ships with Julia .5 and later
+
+if is_unix()
+    provides(Binaries,
+        joinpath(JULIA_HOME, "..", "lib", "julia"),
```
staticfloat: Wow, this *really* should not be necessary.  These libraries should be picked up automatically by the RPATH that is embedded in all Julia builds to find things like `sys.so`.  Also, this will break if someone is running Julia with relative paths to libraries that aren't `../lib/julia`, such as the Debian package where it's `../lib/x86_64-linux-gnu/julia`, etc....

If you don't provide this path, do things break?

staticfloat:
Can you embed a `Bindeps.debug("MbedTLS") into your buildscript before trying to `Pkg.buildI()`?  That will at least give us an idea of what BinDeps sees.  You could also try running something like `julia -e 'Libdl.dlopen("libmbedtls")'` to see what the error message is that is causing Julia to think it doesn't have an `libmbedtls` laying around that it can use.

quinnj:
With julia-provided shared library, do we really need BinDeps anymore? Is that just to provide fallbacks? I'm wondering if we can't just simplify the whole process here and say `const mbedlib = "libmbedtls"` and be done?

malmaud:
Does that actually work right now?

quinnj:
```julia
              _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.5.0 (2016-09-19 18:14 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-apple-darwin13.4.0

julia> data = Libc.malloc(32)
Ptr{Void} @0x00007f8aea45d330

julia> ccall((:mbedtls_pk_init, "libmbedcrypto"), Void, (Ptr{Void},), data)

julia>
```
I mean, 0.5-only I would imagine, but we could bump the Julia minimum.

malmaud:
What do you think, @tkelman ? Any reason not to do this?

staticfloat:
I'm in favor of using the julia-bundled libraries as much as we can, with the caveat that of course we always want to shrink Base as much as possible, so MbedTLS may go away some time in the future.

@quinnj to make sure that it's picking up the right `libmbedcrypto`, you should probably `filter( x-> contains(x, "mbed"), Libdl.dllist())` after that.  For my from-source build of master, things look good:
```
julia> filter( x-> contains(x, "mbed"), Libdl.dllist())
3-element Array{AbstractString,1}:
 "/Users/sabae/src/julia/usr/lib//libmbedtls.dylib"
 "/Users/sabae/src/julia/usr/lib//libmbedx509.dylib"
 "/Users/sabae/src/julia/usr/lib//libmbedcrypto.dylib"
```

quinnj:
Implemented in https://github.com/JuliaWeb/MbedTLS.jl/pull/81

PR #74, opened 2016-10-16
AptGet provider

simonbyrne:


malmaud:
This was already discussed https://github.com/JuliaWeb/MbedTLS.jl/pull/60. The current thinking is to just fix https://github.com/JuliaWeb/MbedTLS.jl/pull/66 and use the mbedtls that ships with julia .5.

simonbyrne:
That makes sense, I didn't realise it was already bundled. Unfortunately I don't really understand BinDeps, so can't help with that.

PR #76, opened 2016-11-16
PK Verification

joshbode:
- added wrapper for `mbedtls_pk_verify`
- inline `sign` function (creates output buffer sized on the private key length)
- inline `parse_public_keyfile` function (ala `parse_keyfile`)
- more tests


quinnj:
Thanks!

joshbode:
Thanks for that! Will it be possible to tag a release off this? Is there anything you need me to do?

PR #78, opened 2016-11-24
Fixed build on ARM platform

tehrengruber:
On ARM devices the build step fails with
```
[  5%] Building C object library/CMakeFiles/mbedcrypto.dir/bignum.c.o
/root/.julia/v0.5/MbedTLS/deps/src/mbedtls-2.1.1/library/bignum.c: In function ‘mpi_mul_hlp’:
/root/.julia/v0.5/MbedTLS/deps/src/mbedtls-2.1.1/library/bignum.c:1128:1: error: r7 cannot be used in asm here
 }
 ^
```

Additional information can be found here: https://tls.mbed.org/kb/development/arm-thumb-error-r7-cannot-be-used-in-asm-here

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -45,7 +45,7 @@ if is_unix()
             @build_steps begin
                 ChangeDirectory(mbed_dir)
                  @build_steps begin
-                    `cmake -DUSE_SHARED_MBEDTLS_LIBRARY=On .`
+                    `cmake -DUSE_SHARED_MBEDTLS_LIBRARY=On $(Sys.ARCH == :arm ? "-DCMAKE_C_FLAGS=-fomit-frame-pointer" : "") .`
                     `make lib`
                 end
             end
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -45,7 +45,7 @@ if is_unix()
             @build_steps begin
                 ChangeDirectory(mbed_dir)
                  @build_steps begin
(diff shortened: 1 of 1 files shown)


yuyichao:
This is an option we usually don't want to include but it shouldn't be terrible since exidx unwinding should be working on master.

This sounds like a compiler bug, any idea which versions of compilers have this issue?

`Sys.ARCH === :arm` is also not enough to detect all arm archs, you need to include basically everything that starts with `arm`.

tehrengruber:
Well this is not really I compiler bug I would say, but more a design decision in MbedTLS. They have optimized assembly code for ARM which uses the same register `r7` as GCC uses for the frame pointer.

> The assembly code in bn_mul.h is optimized for the ARM platform and uses some registers, including r7 to efficiently do an operation. GCC also uses r7 as the frame pointer under ARM Thumb assembly.

There might be other options like disabling the optimized assembly code by some precompiler definitions, but that would require digging into the code and using `-fomit-frame-pointer` is the proposed solution according to MbedTLS. I'm using GCC 5.4, but I would assume that this is not a version specific problem.

```
$ gcc --version
gcc (Ubuntu/Linaro 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609
```

I only have a single arm machine as a reference here, but I can certainly change the if statement to match all architectures which start with `arm`.

To be honest I don't know why unwinding is relevant here.

quinnj:
Shouldn't be needed after https://github.com/JuliaWeb/MbedTLS.jl/pull/81; we're now using the julia-shipped mbed shared libraries.

tehrengruber:
#81 indeed solves this so this can be closed.

PR #81, opened 2016-12-01
Build and 0.5 Update

quinnj:
This updates the package to use the julia-shipped mbed shared libraries, thus avoiding the need for BinDeps.

Tested locally on OSX and Windows.


quinnj:
Passes appveyor & travis on linux.

.travis.yml
```diff
@@ -4,7 +4,7 @@ os:
   - linux
   - osx
 julia:
-  - 0.5
+  - release
```
tkelman: please put this back, release will change over time and is not what you want here

quinnj: Hmm.....yeah, you're probably right. Sorry, I was thinking it would be nicer to stay with whatever the current release is, but it's better to stick to the actual versions that have been tested/vetted against.

tkelman: I want to deprecate `release` for this reason - there's no equivalent in REQUIRE, and it's not good to stop testing supported versions. When you want to test against a new version, best to just add a new line.

quinnj: That's a good idea. It's too tempting at first glance to just put `release` and `nightly`.

joshbode:
To get this working on OS X I needed to install MbedTLS binaries first, e.g.
```bash
$ brew install mbedtls
```

quinnj:
Huh? Are you on Julia 0.5? Julia itself comes with MbedTLS shared libraries, so no additional installation should be needed.

joshbode:
I'm using @staticfloat's homebrew tap for Julia 0.5, e.g.
```bash
$ brew tap staticfloat/julia
...
$ brew install julia
...
$ find /usr/local/opt/julia/lib -name '*mbed*' | wc -l
0
```

joshbode:
The official binaries _do_ contain the libraries:
```bash
$ find ~/Applications/Julia-0.5.app/Contents/Resources/julia/lib -name 'libmbed*' | wc -l
9
```

quinnj:
I wonder if it's just a location issue, @staticfloat, can you say if and where those shared libraries are getting put from the brew tap?

staticfloat:
@quinnj So it turns out that Homebrew's `libgit2` doesn't build against `mbedtls` by default, and since we compile Julia with `USE_SYSTEM_LIBGIT2=1` on Homebrew, we don't get a `libmbedtls.dylib` for free.  Since this is something we've kinda-sorta guaranteed with our "official" builds, I've gone ahead and changed the Julia formula to `depend_on "mbedtls"`.

quinnj:
Thanks @staticfloat.

tkelman:
This isn't going to work with linux distributions' packages of Julia either. This library isn't an absolute dependency of Julia itself, you're going to have to check for it here and have a fallback for when it isn't present.

src/ssl.jl
```diff
@@ -264,5 +264,5 @@ function __sslinit__()
     global const c_recv = cfunction(f_recv, Cint, (Ptr{Void}, Ptr{UInt8}, Csize_t))
     global const c_dbg = cfunction(f_dbg, Void,
         (Ptr{Void}, Cint, Ptr{UInt8}, Cint, Ptr{UInt8}))
-    global const TRUSTED_CERT_FILE = joinpath(dirname(@__FILE__), "../deps/cacert.pem")
```
yuyichao: It seems pretty weird to have `src` reference `test`, why does the file has to be moved?

.travis.yml
tkelman: bump, this should still be changed back

PR #88, opened 2016-12-08
add missing macro hygiene to `err_check` macro

vtjnash:


tkelman:
before tagging this it should be backported to a branch that does not have https://github.com/JuliaWeb/MbedTLS.jl/pull/81. that should not be released as it would brrak the package for anyone using Julia from a package manager.

PR #90, opened 2016-12-16
Add a fallback build.jl step in case correct system libraries aren't detected

quinnj:


deps/build.jl
```diff
@@ -19,9 +19,21 @@ if Libdl.dlopen_e("libmbedtls")    != C_NULL &&
...
+const MBED_TLS = "libmbedtls"
+const MBED_CRYPTO = "libmbedcrypto"
+const MBED_X509 = "libmbedx509"
+""")
+   end
    exit()
 end
 
+println("Manual build...")
+isfile("deps.jl") && rm("deps.jl")
```
tkelman: this doesn't need to be deleted every time, this will cause unnecessary recompiles - only if the content is stale

quinnj: Hmmm.....how do you think I should check for that?  How does BinDeps usually check/know that it needs to rebuild?

tkelman: it compares the content of an existing file (if there is one) and doesn't overwrite if it matches. here was want to be sure we don't use a stale deps.jl if you switch between modes, but avoid recompiling when not necessary if the files don't change while staying in the same mode

quinnj: > it compares the content of an existing file (if there is one) and doesn't overwrite if it matches.
Can you point me to another package where they're doing this? Should be pretty easy to put in if I can follow an example.

tkelman: it's implemented inside bindeps right before it writes the deps.jl file. there are a few packages that may do something similar with non-bindeps manual tag files, possibly Blosc.jl?

yuyichao:
Will this work with 0.4? And will this be merged today?

quinnj:
Does this need to work on 0.4? I can get it wrapped up today.

yuyichao:
> Does this need to work on 0.4?

It doesn't have to, but it seems that it could.

quinnj:
Travis seems happy with the latest commit; any other comments or concerns?

deps/build.jl
```diff
@@ -0,0 +1,100 @@
+function validate_mbed(name, handle)
+    try
+        get_version = Libdl.dlsym(handle, :mbedtls_version_get_string)
+        version_ptr = Vector{UInt8}(9)
+        ccall(get_version, Void, (Ptr{Void},), version_ptr)
+        version = VersionNumber(unsafe_string(pointer(version_ptr)))
+        version >= v"2.1.1"
```
tkelman: does the package work with newer versions too? what about 2.4.0 which currently causes issues with base julia but may be upgraded to later? arch and homebrew may already be using that version

deps/build.jl
```diff
@@ -0,0 +1,100 @@
...
+"""
+
+if Libdl.dlopen_e("libmbedtls")    != C_NULL &&
+   Libdl.dlopen_e("libmbedcrypto") != C_NULL &&
+   Libdl.dlopen_e("libmbedx509")   != C_NULL &&
+   validate_mbed("", Libdl.dlopen_e("libmbedcrypto")) &&
+   get(ENV, "FORCE_BUILD", "") != "true"
+   println("Using system libraries...")
+   open("deps.jl", "w") do f
+       write(f, systemlibs)
```
tkelman: should also only write if the file doesn't exist or doesn't match this content - otherwise changing the timestamp will trigger an unnecessary recompile

quinnj: Good call; updated.

deps/build.jl
quinnj: I thought about updating, but I also didn't want to disrupt too much in this PR. Happy to test out newer versions in a subsequent PR.

tkelman: not asking you to update here. asking if this needs to be more picky - if it's going to pick up system copies of the library, they should really be versions that work

quinnj: Ah, got it. Hmmm, not sure. I can do some testing.

yuyichao:
> And will this be merged today?

So the answer appears to be no.

Does this need another round of review or can this be merged and tagged in a few hours? I do think this is the right way forward but I don't think it's absolutely needed now. Unless this PR is merged before ~ noon tomorrow I'm going to revert #81 and tag it.

tkelman:
If this is more likely than the most recent tag to pick up and use system copies of the library, then it should have a bit more testing to see whether or not the validation check needs to be more picky. If it's equally likely to use a system copy of the library than the most recent tag, then I suspect this is okay for now but the validation testing should still be done as soon as possible.

deps/build.jl
```diff
@@ -0,0 +1,102 @@
...
+                    `make lib`
+                end
+            end
+        end), mbed_all, installed_libpath=joinpath(mbed_dir, "library"))
+end
+
+if is_apple()
+    if Pkg.installed("Homebrew") === nothing
+		error("Homebrew package not installed, please run Pkg.add(\"Homebrew\")")
+	end
```
tkelman: these 3 lines are actually not necessary, people have just been cargo-cult copy pasting them - maybe from some day before `@osx Homebrew` was present in REQUIRE?

yuyichao:
> If this is more likely than the most recent tag to pick up and use system copies of the library

I can at least confirm that the old version can pick up the system version on ArchLinux and that's how I use it. Any other way to tell if the system version is more or less likely to be picked up?

tkelman:
If BinDeps is doing anything more careful than dlopen, I guess. That concerns me since Arch may upgrade to a version that wouldn't work with the package, but if BinDeps also has that problem then it isn't new.

yuyichao:
I don't think BinDeps was doing anything fancy. Can this be merged now?

vtjnash:
Is this ready for a new version tag now? If you enable [AttoBot](https://discourse.julialang.org/t/attobot-a-package-release-bot/494) on the repo, that can be as simple as making a new release tag in Github.

PR #95, opened 2017-01-10
add `set_dbg_level`, do not use `String.data`

### LINKED ISSUE #94: needs update for string.data changes
After JuliaLang/julia#19449, soon to be merged for Julia 0.6, you will no longer be able to access the raw bytes of a string via `string.data`; instead, do `Vector{UInt8}(string)`.  For example, this affects:

https://github.com/JuliaWeb/MbedTLS.jl/blob/d77b6d1ed576f99ea3adbcbcec2c359d692038bf/src/cipher.jl#L195

(Your `tobytes(x)` function could be replaced by `Vector{UInt8}(x)`, it looks like.)

tanmaykm:
- added a new method `set_dbg_level` to manipulate MbedTLS debug threshold
- do not use `String.data`, fixes #94
- fix a deprecation warning in `f_dbg` method

### DIFF
--- a/src/cipher.jl
+++ b/src/cipher.jl
@@ -192,7 +192,7 @@ function get_key_bitlen(cipher::Cipher)
 end
 
 tobytes(x::Vector{UInt8}) = x
-tobytes(x) = String(x).data
+tobytes(x) = Vector{UInt8}(x)
 
 function set_key!(cipher::Cipher, key, op::Operation)
     key_b = tobytes(key)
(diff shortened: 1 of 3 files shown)


tanmaykm:
Failure on nightlies seem to be https://github.com/JuliaLang/julia/issues/19774

quinnj:
@tanmaykm, what would you recommend? merge now? or wait for some resolution to JuliaLang/julia#19774?

tanmaykm:
@quinnj I would prefer to have this merged and create a new issue to track https://github.com/JuliaLang/julia/issues/19774

PR #96, opened 2017-01-16
Use a goto instead of exit() to fix #93

### LINKED ISSUE #93: Running Pkg.build() with MbedTLS fails to build subsequent packages.
I believe this is due to the `exit()` call in the `build.jl`. Pkg.build() exits silently after starting to build MbedTLS without building subsequent packages. Setting `FORCE_BUILD` to `true` enables the build to complete.

tkelman:
closes #93

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -38,14 +38,13 @@ if Libdl.dlopen_e("libmbedtls")    != C_NULL &&
 end
 
 using BinDeps
+@BinDeps.setup
 
 if need_to_build_manually
     println("Manual build...")
     # If we somehow already have a deps from system-libraries, but are now manually building
     # make sure we delete the old deps.jl
     isfile("deps.jl") && readstring("deps.jl") == systemlibs && rm("deps.jl")
-
-    @BinDeps.setup
 
     mbed = library_dependency("libmbedtls", aliases=["libmbedtls", "libmbedtls.2.1.1"])
     mbed_crypto = library_dependency("libmbedcrypto", aliases=["libmbedcrypto", "libmbedcrypto.2.1.1"], validate=validate_mbed)
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -37,13 +37,14 @@ if Libdl.dlopen_e("libmbedtls")    != C_NULL &&
    need_to_build_manually = false
 end
 
+using BinDeps
+
 if need_to_build_manually
     println("Manual build...")
     # If we somehow already have a deps from system-libraries, but are now manually building
     # make sure we delete the old deps.jl
     isfile("deps.jl") && readstring("deps.jl") == systemlibs && rm("deps.jl")
 
-    using BinDeps
     @BinDeps.setup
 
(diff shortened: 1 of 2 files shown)


quinnj:
Can you remind me why we need this?

tkelman:
Because exiting the process prevents any other pending packages from doing anything. Did you read the issue? Or https://github.com/JuliaLang/julia/issues/20020.

quinnj:
Ah that's right. Thanks for the reminder.

Hmmmm....can you use `@goto` at the toplevel? Appveyor doesn't seem to like what's going on.

tkelman:
I will admit I didn't actually test this, was hoping it would work but apparently not.

tkelman:
I give up for now. This needs fixing ASAP, but I have to do other things.

quinnj:
@yuyichao, do you know why we'd be seeing there errors [here](https://ci.appveyor.com/project/quinnj/mbedtls-jl/build/job/nodolkwrtj6n7owf) on 0.6 vs. 0.5 passing?

yuyichao:
Fix https://github.com/JuliaWeb/MbedTLS.jl/blob/d77b6d1ed576f99ea3adbcbcec2c359d692038bf/src/cipher.jl#L195 ?

quinnj:
But the error we're seeing is
```julia
LoadError: error compiling anonymous: error in method definition: function BinDeps.library_dependency must be explicitly imported to be extended
```

yuyichao:
Minimum repro is

```jl
julia> module M
       export f
       f() = 1
       end
M

julia> using M

julia> begin
           f() = 1
           f()
       end
ERROR: error in method definition: function M.f must be explicitly imported to be extended
```
I get the same behavior on 0.3 0.4 and 0.5 though.

quinnj:
But I don't think we're even trying to define a method of `library_dependency` anywhere (that I'm aware of); we're only trying to *call* it.

```julia
    mbed = library_dependency("libmbedtls", aliases=["libmbedtls", "libmbedtls.2.1.1"])
    mbed_crypto = library_dependency("libmbedcrypto", aliases=["libmbedcrypto", "libmbedcrypto.2.1.1"], validate=validate_mbed)
    mbed_x509 = library_dependency("libmbedx509", aliases=["libmbedx509", "libmbedx509.2.1.1"])
```

yuyichao:
It comes from `BinDeps.@setup`. You should be able to workaround it by moving it out of the `if`.

tkelman:
Thanks guys. At least this gets through the bindeps part on appveyor.

quinnj:
I can't do anything with this for the next couple of hours, but I just pushed a fix for the `.data` usage in `cipher`. If someone wants to track travis/appveyor and merge on green, that'd be great.

quinnj:
Merging this as the 0.6 errors are back to what they were (which, IIUC, depend on JuliaLang/julia#19774)

tkelman:
Thanks Jacob. Tagging this.

PR #99, opened 2017-03-15
Update dep takebuf_array method

musm:
Please tag after a merge, thanks

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -224,7 +224,7 @@ function Base.readavailable(ctx::SSLContext)
     b = IOBuffer()
     write(b, read(ctx, 1))
     write(b, read(ctx, nb_available(ctx)))
-    return takebuf_array(b)
+    return take!(b)
 end
 
 function Base.eof(ctx::SSLContext)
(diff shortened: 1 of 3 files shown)


src/ssl.jl
```diff
@@ -224,7 +224,7 @@ function Base.readavailable(ctx::SSLContext)
     b = IOBuffer()
     write(b, read(ctx, 1))
     write(b, read(ctx, nb_available(ctx)))
-    return takebuf_array(b)
+    return take!(b)
```
malmaud: Shouldn't it be `@compat take!(b)`?

musm: no, see https://github.com/JuliaLang/Compat.jl/blob/master/src/Compat.jl#L1159

malmaud:
Ah, OK.

src/ssl.jl
tkelman: you do need to use or import Compat though

musm:
g2g

musm:
Is this ok? Thx

PR #106, opened 2017-05-17
Fix wrong use of reinterpret in tests

yuyichao:


yuyichao:
Ref https://github.com/JuliaLang/julia/pull/21831

PR #111, opened 2017-08-17
Try out dropping Compat

quinnj:
compat


src/cipher.jl
```diff
@@ -275,7 +275,7 @@ end
 crypt!(cipher::Cipher, buf_in, buf_out) = crypt!(cipher, C_NULL, buf_in, buf_out)
 
 function crypt(cipher_info, op::Operation, key, iv, msg)
-    const max_block_size = 256  # todo: obtain this dynamically from mbedtls
+    max_block_size = 256  # todo: obtain this dynamically from mbedtls
```
quinnj: @keno, do you think femtocleaner could learn this one? (`const` in a local block)

Keno: Yes, that's an easy one

PR #114, opened 2017-12-13
Fix #113 - empty read after eof() false

### LINKED ISSUE #113: readavailable() = [] after eof() = false
I'm debugging some code in HTTP.jl where it is assumed  to be [_"always safe to read one byte after seeing eof return false"_](https://docs.julialang.org/en/stable/stdlib/io-network/#Base.eof).

The loop looks like this:
```julia
    while !eof(conn.socket)
        bytes = readavailable(conn.socket)
        @assert length(bytes) > 0
        ...
    end
```

When I do `HTTP.post("https://httpbin.org/post"; body="hey", chunksize=2))`, the assertion fails.

As a workaround I have this:
```julia
    while !eof(conn.socket)
        bytes = readavailable(conn.socket)
        if length(bytes) == 0
            @assert isa(conn.socket, MbedTLS.SSLContext)
            @assert eof(conn.socket)
            break
        end
        @assert length(bytes) > 0
        ...
    end
```
i.e. it looks like a read is needed in order for `eof` to notice that the socket is closed ?

Note: the code has comments about `eof` and `nb_available` oddness. Presumably this is related.

https://github.com/JuliaWeb/MbedTLS.jl/blob/43e1a4552063566761ea72bce47db0aee44666c8/src/ssl.jl#L223

https://github.com/JuliaWeb/MbedTLS.jl/blob/43e1a4552063566761ea72bce47db0aee44666c8/src/ssl.jl#L212

samoconnor:
Attempted fix for #113

See https://esp32.com/viewtopic.php?t=1101#p4884

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -208,23 +208,11 @@ function Base.readbytes!(ctx::SSLContext, buf::Vector{UInt8}, nbytes::UInt)
     return Int(nr::UInt)
 end
 
-function Base.readavailable(ctx::SSLContext)
-    # For unknown reasons, nb_available on SSLContext erroneously returns 0
-    # until `read` is called on the context. As a temporary hack, we read one
-    # byte from the SSL context to cause nb_available to be accurate.
-    # TODO: figure out and fix the root cause of this
-    b = IOBuffer()
-    write(b, read(ctx, 1))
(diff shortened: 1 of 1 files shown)


samoconnor:
@yuyichao, @quinnj please review

PR #118, opened 2018-01-05
0.7 compat

quinnj:


src/entropy.jl
```diff
@@ -15,29 +15,28 @@ mutable struct Entropy
...
     sz = f(output_jl)
     unsafe_store!(convert(Ptr{Csize_t}, olen), Csize_t(sz))
     return Cint(0)
 end
 
-function add_source!(ctx::Entropy, f, threshold, strong)
+function add_source!(ctx::Entropy, f::Function, threshold, strong)
     push!(ctx.sources, f)
-    add_source!(ctx, c_entropy[], pointer_from_objref(f), threshold, strong ? 1 : 0)
+    add_source!(ctx, c_entropy[], f, threshold, strong ? 1 : 0)
```
quinnj: @yuyichao @Keno, is this a bad idea to just pass Julia objects as `Any` in C callbacks? i.e. up in `add_source!` we pass a `Function` as `Any` and then in `jl_entropy` we just call the `f::Function` directly. All the tests seem to be happy locally.

PR #119, opened 2018-01-15
Explicitly import Libdl

staticfloat:
https://github.com/JuliaLang/julia/pull/25459 moved `Libdl` to the stdlib, which means that we need to explicitly import it now before we can do things like `Libdl.dlopen()`.

Note that this is essentially a no-op on Julia 0.6, so this shouldn't break anything.  (We'll see if CI agrees with me there)

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,4 +1,7 @@
 using Compat
+if VERSION >= v"0.7.0-DEV.3382"
+    import Libdl
+end
 
 need_to_build_manually = true
 


staticfloat:
......okay apparently nightlies are a bit behind.  I guess we'll wait on this until I fix the nightlies.

staticfloat:
Okay great, these are passing now, and it's just https://github.com/JuliaWeb/MbedTLS.jl/issues/120 killing things on 0.7.

quinnj:
Thanks for the PR! I fixed the other issue; so hopefully master is good now again.

PR #123, opened 2018-02-27
Jq/bindeps2

quinnj:


quinnj:
@ararslan @yuyichao @Keno @malmaud @staticfloat, any body have any concerns w/ doing this? I noticed today that MbedTLS.jl is actually the only dependency of BinaryProvider/Builder that uses BinDeps.jl, so making this change allows "BinDeps2" to not have an upstream dependency on BinDeps.jl!

ararslan:
I'm against dropping BinDeps until there's a proper solution for platforms like FreeBSD and Alpine which support Julia but are not supported by BinaryBuilder/Provider.

REQUIRE
```diff
@@ -1,4 +1,3 @@
 julia 0.6
-BinDeps
-@osx Homebrew
+BinaryProvider
 Compat 0.41.0
```
tkelman: 0.48 for Libdl

quinnj:
So.......I think this might actually work on those platforms; there's nothing platform specific (I think) about how we're using BinaryProvider here. Can you test on any of those? I think I can spin up an alpine docker container, though I wonder if julia even runs on alpine.

tkelman:
The problem for this, and why the BinDeps fallback was needed here, was to handle non-default build configurations of julia that may not always bundle a copy of mbedtls. If your distro happens to have a close-enough version of libgit2 that links against openssl or whatever else for https support, you can build a fully functional julia against it and it's not valid to assume mbedtls is present. The fedora package of julia may be a case of this, though @nalimilan can correct me if a more obscure example is needed for what would break from this.

nalimilan:
Yes, the Fedora package (as most distribution will probably do) uses the system libgit2, which is linked to OpenSSL. I'm happy to run tests with that build if that's useful.

nalimilan:
Yes, the Fedora package (as most distribution will probably do) uses the system libgit2, which is linked to OpenSSL. I'm happy to run tests with that build if that's useful.

quinnj:
@nalimilan, so does that mean the actual julia package will be built w/ `USE_SYSTEM_MBED` on Fedora? Like, there definitely won't be the three mbed shared libraries that typically get built and shipped w/ julia?

nalimilan:
> @nalimilan, so does that mean the actual julia package will be built w/ USE_SYSTEM_MBED on Fedora? Like, there definitely won't be the three mbed shared libraries that typically get built and shipped w/ julia?

Right.

tkelman:
More than that though, it's system libgit2 with no mbedtls at all.

quinnj:
Alright, update here that will hopefully satisfy everyone (and still avoid the BinDeps.jl dependency). We now employ 3 strategies for satisfying MbedTLS.jl binary products:
1. Use pre-built binaries built by BinaryBuilder, repo [here](https://github.com/quinnj/MbedTLSBuilder); this has the advantage of not tying the majority of users (on supported platforms) to julia's own mbedtls shipped binaries, which could be cause for concern if there was an urgent vulnerability found and needed to be patched; it also allows the package to update separately from Base julia anyway
2. Use the julia-shipped shared libraries; this is a great fallback, because it means if there's a system that we build julia for, they can still be supported via the shipped libraries
3. Build from source; for cases where users are using custom distros or for some reason don't have julia-shipped shared libraries, we can just build from source, which is pretty straightforward and one of the easiest builds I've encountered (at least w/ my experience going through the BinaryBuilder process)

That's about as good as we can get and probably even more than is expected.

deps/build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+# meant to be run from deps/ directory to satisfy build.jl
+if ! cmake --version > /dev/null 2>&1; then
+  echo "cmake command not found. installing cmake is required for building from source."
+  exit 1
+fi
+
+git clone https://github.com/ARMmbed/mbedtls.git
+VERSION="2.7.1"
+git checkout mbedtls-$VERSION
```
tkelman: download a tarball, not a git clone, respecting the license choice as build.jl was doing

.travis.yml
```diff
@@ -1,8 +1,5 @@
 # Documentation: http://docs.travis-ci.com/user/languages/julia/
 language: julia
-env:
-  - FORCE_BUILD=true
```
tkelman: should keep testing these

deps/build.sh
quinnj: Good call; I was trying to make sure I got everything ported over, but forgot the gpl vs. apache.

deps/build.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+# meant to be run from deps/ directory to satisfy build.jl
+if ! cmake --version > /dev/null 2>&1; then
+  echo "cmake command not found. installing cmake is required for building from source."
+  exit 1
+fi
+
+VERSION="2.7.0"
+
+if [ -z "$" ]; then
```
tkelman: USE_GPL_MBEDTLS

.travis.yml
```diff
@@ -1,8 +1,8 @@
...
-  - FORCE_BUILD=false
+-env:
```
tkelman: no leading dash, I'm surprised travis didn't fail on this

deps/build.jl
```diff
@@ -1,111 +1,64 @@
...
+    BinaryProvider.Linux(:x86_64, :glibc) => ("$bin_prefix/MbedTLS.x86_64-linux-gnu.tar.gz", "1603eb503252e7bad78691e81230de5f4bbaf1a369fdf8812a9b9e9a16ccf321"),
+    BinaryProvider.Windows(:x86_64) => ("$bin_prefix/MbedTLS.x86_64-w64-mingw32.tar.gz", "0a9a6f05dc08349294934e412836bdd292175883d9df015d93d19d5da7429d9d"),
+)
 
-    if haskey(ENV, "USE_GPL_MBEDTLS")  # The source code is identical except for the license text
-        source_uri = URI("https://cache.julialang.org/https://tls.mbed.org/download/mbedtls-2.1.1-gpl.tgz")
-        srcsha = "22c76e9d8036a76e01906423b3e8a02ab0ef84027f791bd719fff8edee9c61a9"
+# First, check to see if we're all satisfied
+if any(!satisfied(p; verbose=verbose) for p in products) || get(ENV, "FORCE_BUILD", false)
+    if haskey(download_info, platform_key()) && get(ENV, "FORCE_BUILD", "false") != "true"
```
staticfloat: Let's open an issue on BinaryBuilder to maybe do this by default in our generated `build.jl` files; I'd like to standardize this across packages so that we don't have a bunch of random environment variables that are used differently by different packages that need to do this.

quinnj:
@tkelman, any idea why the source build failed to find `cmake` on travis? From what I can tell, it should just be available by default
```
NFO: Downloading https://tls.mbed.org/download/mbedtls-2.7.0-apache.tgz to /tmp/juliaroZkcG-download.tar.gz...
[20:37:26] ######################################################################## 100.0%
INFO: No hash cache found
INFO: Calculated hash aeb66d6cd43aa1c79c145d15845c655627a7fc30d624148aaafbb6c36d7f55ef for file /tmp/juliaroZkcG-download.tar.gz
INFO: Destination directory /home/travis/.julia/v0.6/MbedTLS/deps already exists, returning
cmake command not found. installing cmake is required for building from source.
===============================[ ERROR: MbedTLS ]===============================
```
full logs: https://travis-ci.org/JuliaWeb/MbedTLS.jl/jobs/348407488

tkelman:
the `env` kwarg to run replaces all environment variables, it doesn't append. need to do something like
```
ENV2 = copy(ENV)
ENV2["VERSION"] = ...
```
and pass `ENV2`

deps/build.jl
```diff
@@ -1,111 +1,67 @@
...
+        url, hash = haskey(ENV, "USE_GPL_MBEDTLS") ?
+            ("https://tls.mbed.org/download/mbedtls-$VERSION-gpl.tgz", "2c6fe289b4b50bf67b4839e81b07fcf52a19f5129d0241d2aa4d49cb1ef11e4f") :
+            ("https://tls.mbed.org/download/mbedtls-$VERSION-apache.tgz", "aeb66d6cd43aa1c79c145d15845c655627a7fc30d624148aaafbb6c36d7f55ef")
+        download_verify(url, hash, joinpath(@__DIR__, "mbedtls.tgz"), force=true, verbose=true)
+        unpack(joinpath(@__DIR__, "mbedtls.tgz"), @__DIR__; verbose=true)
+        withenv("VERSION"=>VERSION) do
+            run(Cmd(`./build.sh`, dir=@__DIR__))
+        end
+        if any(!satisfied(p; verbose=verbose) for p in products)
+            error("attmped to build mbedtls shared libraries, but they couldn't be located (deps/usr/lib)")
```
tkelman: attempted

deps/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
tkelman: should add a `set -e` here so the script fails early if any step fails

src/MbedTLS.jl
```diff
@@ -51,7 +41,12 @@ export
 
 import Base: show
 
-include(joinpath(dirname(@__FILE__),"..","deps","deps.jl"))
+# Load libmariadb from our deps.jl
```
tkelman: libmbedtls

.travis.yml
```diff
@@ -1,8 +1,16 @@
 # Documentation: http://docs.travis-ci.com/user/languages/julia/
 language: julia
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - gcc-6
+      - g++-6
+      - cmake
```
tkelman: these shouldn't be needed

deps/build.jl
quinnj: what exactly do you want to generalize? building from source as a backup? or just the environment variable I'm using in travis here?

tkelman:
Yeah, this looks like a reasonable no-worse-than-what-was-there-before version, assuming `BinaryProvider.satisfied` can be fully trusted to identify when its binaries won't work before trying to use them. As commented on Ipopt.jl I don't think BinaryBuilder's products will always work correctly with source builds of Julia, but I don't know whether it's smart enough to identify problematic cases early enough to correctly trigger the source-build fallback. Keep an eye on it.

Which license version of mbedtls did you use to make the BinaryProvider builds? might need to also switch those off when the opposite license is asked for by the user.

quinnj:
I just built straight from the mbedtls github repo (checking out latest release tag).

tkelman:
I think it's apache in the repo so if someone sets USE_GPL_MBEDTLS they probably don't want to be using apache-licensed binaries

deps/build.jl
staticfloat: The environment variable.  :)

tkelman:
Again, if the user asks for gpl binaries, the apache dbinaries should not get used.

and this should preserve the validation version check, otherwise some of these paths could pick up wrong-version system copies

quinnj:
Ah yes, I'll add that back in.

Also, I would now argue that we're doing _better_ than before because we're now automatically providing an _up-to-date_ version of the libraries and not tying ourselves to whatever ships w/ julia. This is both for prebuild binaries and source builds.

tkelman:
theoretically, if the binarybuilder products work in cases that the bundled mbedtls doesn't. I don't think it's been demonstrated that binarybuilder will work well across varied source build configurations of julia. at least there's no fortran here which is the most obvious problem, but I suspect libgcc and libstdc++ abi incompatibilities will start popping up

quinnj:
Addressed: d1a9025a78b191ba30941b6ec28fe1a1a83ff6cf

tkelman:
partially, thanks. still need to put back the version validation check to avoid picking up too-old system copies

PR #124, opened 2018-03-02
Try to fix deadlock

samoconnor:
This call:
nb_available(::SSLContext)->mbedtls_ssl_read->f_recv->readbytes!()
can block holding the datalock and can stop calls write(::SSLContext) calls
(in other tasks) from getting the lock and writing. If a request is not
writen, the response may never come, and the read may block forever.

The forum post below says that the zero-size mbedtls_ssl_read call
used to find out how many bytes are available must be done in
non-blocking mode. This commit addes a nonblocking flag to try to
implement that.

This seems to work to avoid a deadlock in HTTP.jl where:
 - A request header has been written.
 - An async task has been schedueled to write the body.
 - The main task starts reading the response (in some cases the
   response headers arrive before the request body is sent).

The expected behaviour is that the reading task will soon block to wait
for data, thereby yeilding to the async task, which will finish writing
the body. Before this commit, it would get stuck.

See: https://esp32.com/viewtopic.php?t=1101#p4945
    "After the correct implementation of non-blocking sockets, the things
    seem to be working fine"

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -24,6 +24,7 @@ Base.show(io::IO, c::SSLConfig) = print(io, "MbedTLS.SSLConfig()")
 mutable struct SSLContext <: IO
     data::Ptr{Cvoid}
     datalock::ReentrantLock
+    nonblocking::Bool
     config::SSLConfig
     isopen::Bool
     bio
@@ -32,6 +33,7 @@ mutable struct SSLContext <: IO
         ctx = new()
         ctx.data = Libc.malloc(1000)  # 488
         ctx.datalock = ReentrantLock()
+        ctx.nonblocking = false
         ccall((:mbedtls_ssl_init, MBED_TLS), Cvoid, (Ptr{Cvoid},), ctx.data)
         @compat finalizer(ctx->begin
             ccall((:mbedtls_ssl_free, MBED_TLS), Cvoid, (Ptr{Cvoid},), ctx.data)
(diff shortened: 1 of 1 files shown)


samoconnor:
@Keno @quinnj @malmaud

quinnj:
> in some cases the response headers arrive before the request body is sent

Is that just in `100-continue` situations or are there other cases where this can happen?

src/ssl.jl
```diff
@@ -278,8 +284,13 @@ function Base.nb_available(ctx::SSLContext)
     @lockdata ctx begin
         # First try to read from the socket and decrypt incoming data if
         # possible. https://esp32.com/viewtopic.php?t=1101#p4884
-        ccall((:mbedtls_ssl_read, MBED_TLS),
-              Cint, (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t), ctx.data, C_NULL, 0)
+        ctx.nonblocking = true
+        try
```
quinnj: is this try-catch really needed? Can `ccall`s actually throw errors?

samoconnor:
No, it is quite common for the server to immediately respond with error response headers if it can tell it isn't happy based on the request headers. e.g. unsupported request, or bad credentials.

samoconnor:
In fact there is a specific part in the RFC about how if the server says Connection: Close you must close before sending the remainder of the body. There is a test for this in test/loopback.jl

src/ssl.jl
```diff
@@ -112,19 +114,23 @@ end
 function f_send(c_ctx, c_msg, sz)
     jl_ctx = unsafe_pointer_to_objref(c_ctx)
     jl_msg = unsafe_wrap(Array, c_msg, sz)
-    return Cint(write(jl_ctx, jl_msg))
+    return Cint(write(jl_ctx.bio, jl_msg))
```
quinnj: is this right? we're calling `write` on the TCPSocket now instead of on the `SSLContext`?

src/ssl.jl
samoconnor: I'm not 100% sure.
But it's never good for a block of code to appear to restore the old state, but have a lurking possibility of exiting early.
(& It isn't a try/catch, its a try/finally)
I'm happy to leave a FIXME in the code suggesting that someone could take it out in the future if everything is working nicely...

src/ssl.jl
```diff
@@ -112,19 +114,23 @@ end
...
+    else
+        n = readbytes!(jl_ctx.bio, jl_msg, sz)
+    end
     return Cint(n)
 end
 
 function set_bio!(ssl_ctx::SSLContext, jl_ctx::T) where {T<:IO}
     ssl_ctx.bio = jl_ctx
-    set_bio!(ssl_ctx, pointer_from_objref(jl_ctx), c_send[], c_recv[])
+    set_bio!(ssl_ctx, pointer_from_objref(ssl_ctx), c_send[], c_recv[])
```
quinnj: what effect does this change end up having?

src/ssl.jl
samoconnor: Before the void* passed to the library was the socket.
Now it is the SSLContext. This is needed so we can see the nonblocking flag in the callback.
So, the `.bio` is needed now to get back to the socket.

src/ssl.jl
samoconnor: see above

src/ssl.jl
quinnj: Oh, that makes sense. 👍

quinnj:
Looks good. Thanks @samoconnor !

samoconnor:
>> Maybe more generally shouldn't be waiting for exactly sz bytes

>yeah, I actually had code that did just that, but I thought it was best to keep the change small for now.

@vtjnash, @quinnj, see: https://github.com/JuliaWeb/MbedTLS.jl/pull/125

PR #141, opened 2018-03-27
Fix #134

### LINKED ISSUE #134: get(ENV, "...", false) needs quotes
`get(ENV, "FORCE_BUILD", false)` should be `get(ENV, "FORCE_BUILD", "false")` in deps/build.jl

samoconnor:
s2maki commented:
> get(ENV, "FORCE_BUILD", false) should be get(ENV, "FORCE_BUILD", "false") in deps/build.jl

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -33,7 +33,7 @@ download_info = Dict(
 )
 
 # First, check to see if we're all satisfied
-if any(!satisfied(p; verbose=verbose) for p in products) || get(ENV, "FORCE_BUILD", false)
+if any(!satisfied(p; verbose=verbose) for p in products) || get(ENV, "FORCE_BUILD", "false")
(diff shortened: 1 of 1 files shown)


s2maki:
I think it needs more than that.  The `get` returns a string, but it is being used in an `if`.  Something more like
```
get(ENV, "FORCE_BUILD", "false") == "true"
```
as it is in the line below.

samoconnor:
@quinnj looking at this code in more detail I notice:
 - It would probably be best to rename `FORCE_BUILD` -> `MBED_TLS_JL_FORCE_BUILD` and `USE_GPL_MBEDTLS` -> `MBED_TLS_JL_USE_GPL` to avoid possible unintended influence of 3rd party env vars.
 - It might be better to use `haskey(ENV, ...)` (as you do for the GPL flag) for the FORCE flag rather than relying on exact string match of "true" and "false".

PR #142, opened 2018-04-03
Couple of fixes

quinnj:


README.md
```diff
@@ -6,6 +6,8 @@
 
 A wrapper around the [mbed](https://tls.mbed.org/) TLS and cryptography C libary.
 
+MbedTLS.jl currently supports versions 2.7.0/2.7.1 of [ARMmbed/mbedtls](https://github.com/ARMmbed/mbedtls).
+
```
samoconnor: Can we add something like this ? (assuming that it is now true)
```
To ensure that a compatible version of the Mbed TLS library is always used,
MbedTLS.jl does not use local system Mbed TLS binaries.
Instead it installs pre-compiled binaries from [MbedTLSBuilder]
(https://github.com/JuliaWeb/MbedTLSBuilder/releases).
```

src/MbedTLS.jl
```diff
@@ -48,6 +48,13 @@ if !isfile(depsjl_path)
 end
 include(depsjl_path)
 
+const VERSION_BUFFER = zeros(UInt8, 9)
+function getversion()
+    ccall((:mbedtls_version_get_string, libmbedtls), Cvoid, (Ptr{UInt8},), VERSION_BUFFER)
+    return VersionNumber(unsafe_string(pointer(VERSION_BUFFER)))
+end
+const CURRENT_MBEDTLS_VERSION = v"2.7.0"
```
tkelman: what api is being used that first exists in 2.7.0?

tkelman:
this should be an install time validation, not a runtime warning.

src/MbedTLS.jl
quinnj: it's the official version we're installing from MbedTLSBuilder and so is the version that is actively tested against and expected to be installed. If a user is using an older version, it may contain known vulnerabilities which I think is fair to warn about.

quinnj:
I'm not sure it really matters if it's install or runtime; we're just emitting a warning, not throwing an error. Either way, they're going to have to take a look at things and restart julia.

quinnj:
Argh, not sure why the mbedtls version function isn't showing up in the windows libraries.

samoconnor:
Where does this stand now @quinnj ?

README.md
samoconnor: bump

PR #143, opened 2018-04-06
fix missing string comparison in build.jl

rdeits:
Currently the build fails for me with:

```
LoadError: TypeError: non-boolean (String) used in boolean context
while loading /home/rdeits/locomotion/explorations/learning-mpc-2/packages/v0.6/MbedTLS/deps/ starting on line 36
```

because line 36 returns the string "false" instead of a boolean. This PR fixes the error.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -33,7 +33,7 @@ download_info = Dict(
 )
 
 # First, check to see if we're all satisfied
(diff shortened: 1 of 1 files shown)


rdeits:
It looks like this is redundant with one of the changes in #142, so feel free to close it if you'd rather merge that PR instead.

Issue #148, opened 2018-05-11
new handshake changes appear to be throwing ECONNRESET unexpectedly 


s2maki:
When using 1c8c0a1 and trying to make several calls to AWS in quick succession, I am seeing the following error:

```
ERROR (unhandled task failure): read: connection reset by peer (ECONNRESET)
Stacktrace:
 [1] try_yieldto(::Base.##296#297{Task}, ::Task) at ./event.jl:189
 [2] wait() at ./event.jl:234
 [3] wait(::Condition) at ./event.jl:27
 [4] macro expansion at /home/jrun/.julia/v0.6/MbedTLS/src/ssl.jl:188 [inlined]
 [5] (::MbedTLS.##36#37{MbedTLS.SSLContext})() at ./event.jl:73
```

The line in question (ssl.jl:188) appears to be waiting on a Condition (`ctx.bio.readnotify`) , but somehow it ends up throwing an `ECONNRESET` in some cases.

While it does appear from the Julia source that this Condition is allowed to throw exceptions, there is no handler for it since this is happening in a `@schedule`.  Separately, I'm not sure why the `ECONNRESET` is occurring in the first place.  If I roll back to v0.5.8, there are no errors for the same sequence of AWS calls.

This is in Julia 0.6 on Linux.

quinnj:
@samoconnor ^^

samoconnor:
https://github.com/JuliaWeb/MbedTLS.jl/blob/1c8c0a1b8d0e4baa1774a44794447659fd91124b/src/ssl.jl#L183-L190

It seems that the `readnotify` `Condition` can throw a `UVError` in [`uv_readcb (base/stream.jl:485)`](https://github.com/JuliaLang/julia/blob/v0.6.2/base/stream.jl#L485) when libuv signals a fatal connection error to the read callback.

The usual place that waits on `readnotify` is [`wait_readnb`](https://github.com/JuliaLang/julia/blob/v0.6.2/base/stream.jl#L296) (called by the various other Base read functions). My guess is that the `ECONNRESET` was previously showing up as a read error and was handled transparently by the retry loops in HTTP.jl / AWSCore.jl.

The intention of the `@schedule while isopen...` task is to keep the connection active in libuv so that TLS `close_notify` messages can be processed even if there is no other activity on the connection. We're calling `wait(ctx.bio.readnotify)` as a convenient way to sleep most of the time. 
[...]

PR #149, opened 2018-05-15
ignore UVError ECONNRESET in while isopen(ctx) / start_reading loop

samoconnor:
See #148


s2maki:
Can this be reviewed and merged soon?  The currently released version of MbedTLS has troubles with connection hanging, and the head of master has this bug.  It would be nice to have a more stable version.

Thanks!

quinnj:
Let's have CI re-run this just to make sure things are fresh and then we should be good to merge.

PR #158, opened 2018-08-03
Fix remaining julia 0.7 dep warnings

davidanthoff:
I only saw #157 after I had finished this PR, so it does that one as well. This PR here fixes more warnings, though, so we might as well just merge this here.

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -207,8 +207,8 @@ function handshake(ctx::SSLContext)
             try
                 wait(ctx.bio.readnotify)
             catch e
-                if e isa Base.UVError
-                    # Ignore read errors (UVError ECONNRESET)
+                if e isa Base.IOError
+                    # Ignore read errors (IOError ECONNRESET)
                     # https://github.com/JuliaWeb/MbedTLS.jl/issues/148
                 else
                     rethrow(e)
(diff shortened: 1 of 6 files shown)


davidanthoff:
I don't understand why this PR would break things on 0.6, it really doesn't seem to change anything there?

src/ssl.jl
```diff
@@ -207,8 +207,8 @@ function handshake(ctx::SSLContext)
             try
                 wait(ctx.bio.readnotify)
             catch e
-                if e isa Base.UVError
-                    # Ignore read errors (UVError ECONNRESET)
+                if e isa Base.IOError
```
quinnj: this will need to be compat-ified, right?

davidanthoff: I think the whole larger block is already behind a version check.

JeffBezanson:
Can this please be merged? Existing versions will still work on 0.6 so I don't see a major problem there.

tkelman:
It needs to drop 0.6 in REQUIRE if it doesn't work there.

src/ssl.jl
tkelman: if the renaming happened after 0.7.0-alpha.0, then the cutoff should be changed

JeffBezanson:
Yes, you're right about REQUIRE.

PR #159, opened 2018-08-07
Davidanthoff julia 0.7

quinnj:


REQUIRE
```diff
@@ -1,3 +1,3 @@
-julia 0.6
+julia 0.7
```
tkelman: this should say whichever of the rc's was the first that had made the IOError renaming

tkelman: this won't work as a lower bound until 0.7.0 final is released

PR #166, opened 2018-09-20
return MBEDTLS_ERR_NET_CONN_RESET from f_recv when closed by peer

### LINKED ISSUE #164: Stuck in infinite handshake loop when performing a non https connection with correct SSLConfig provided.
**Short description issue**
Stuck in infinite handshake loop when performing a non-https connection such as by `netcat` or http request to `HTTP.listen()` with correct SSLConfig.

**Environment**
```
Ubuntu 18.04 x64
Julia v1.0.0
julia> Pkg.status() 
    Status `~/.julia/environments/v1.0/Project.toml`
  [cd3eb016] HTTP v0.6.14
  [739be429] MbedTLS v0.6.0
```
**Code** 
Given is `server.jl` with `HTTP.listen()` with correct SSLConfig. HTTPS request are handled, so there are no problems with certificate and key.
```julia
using HTTP
using MbedTLS

# SSL configuration via files (cert and key)
const CERT_FILE = "server_cert.pem"
const PK_FILE = "server_key.pem"

println("Create the SSLConfig")
tlsconfig = invoke(
  HTTP.Servers.SSLConfig, Tuple{Any,Any}, CERT_FILE, PK_FILE
)

function show_debug(level, filename, number, msg)
  println("[DBG] $filename $number $msg")  
end

MbedTLS.dbg!(tlsconfig,show_debug)
MbedTLS.set_dbg_level(MbedTLS.INFO)

println(tlsconfig)
println(tlsconfig.cert)

HTTP.listen("0.0.0.0", 8000, ssl=true, sslconfig = tlsconfig) do request::HTTP.Request
   @show request
   @show request.method
   @show HTTP.header(request, "Content-Type")
   try
      return HTTP.Response("Hello Julia World!")
   catch e
      return HTTP.Response(404, "Error: $e")
   end
end
```
**Creation of error situation**
If I run `server.jl` and execute the the following command in a console (where dev.xxxxxxxx.com maps to 127.0.0.1):
```bash
$ nc dev.xxxxxxxx.com 8000
^C
```
this action, `CTRL-C`, results in the following output in the  Julia terminal:
```
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 6754 => handshake
[DBG] /workspace/srcdir/mbedtls/library/ssl_srv.c 4219 server state: 1
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 2471 => flush output
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 2483 <= flush output
[DBG] /workspace/srcdir/mbedtls/library/ssl_srv.c 1192 => parse client hello
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 2252 => fetch input
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 2413 in_left: 0, nb_want: 5
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 2437 in_left: 0, nb_want: 5
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 6764 <= handshake
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 6754 => handshake
[DBG] /workspace/srcdir/mbedtls/library/ssl_srv.c 4219 server state: 1
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 2471 => flush output
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 2483 <= flush output
[DBG] /workspace/srcdir/mbedtls/library/ssl_srv.c 1192 => parse client hello
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 2252 => fetch input
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 2413 in_left: 0, nb_want: 5
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 2437 in_left: 0, nb_want: 5
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 6764 <= handshake
[DBG] /workspace/srcdir/mbedtls/library/ssl_tls.c 6754 => handshake
[DBG] /workspace/srcdir/mbedtls/library/ssl_srv.c 4219 server state:
... infinitely.
```
and I cannot perform any other request, e.g. curl https://dev.xxxxxxxx.com:8000.
When I press CTRL-C `listen()` is interrupted:
```
^C┌ Warning: Interrupted: listen(Sockets.InetAddr{Sockets.IPv4}(ip"0.0.0.0", 0x1f40))
└ @ HTTP.Servers ~/.julia/packages/HTTP/nUK4f/src/Servers.jl:379 
```
**Analysis**
I have performed some analysis and I think the problem is caused by the fact that in `MbedTLS.jl` in file `ssl.jl` [function frecv()](https://github.com/JuliaWeb/MbedTLS.jl/blob/4911ba37607ba1b14a24cb39ae6de886c1d2e97f/src/ssl.jl#L117) does not terminate in case the underlying connection is closed:

```julia
function f_recv(c_ctx, c_msg, sz)
    n = bytesavailable(jl_ctx)
    if n == 0
        return Cint(MBEDTLS_ERR_SSL_WANT_READ)
    end
    n = min(sz, n)
    unsafe_read(jl_ctx, c_msg, n)
    return Cint(n)
end
```
When I look at the function int [mbedtls_net_recv( void *ctx, unsigned char *buf, size_t len )](https://github.com/ARMmbed/mbedtls/blob/8266acacc8d6e1c65fba9a048f56339d0827b2fe/library/net_sockets.c#L535) as used by example [ssl_server.c](https://github.com/ARMmbed/mbedtls/blob/mbedtls-2.11.0/programs/ssl/ssl_server.c) code for MbedTLS 2.11.0 I see the following code:

```c
        if( errno == EPIPE || errno == ECONNRESET )
            return( MBEDTLS_ERR_NET_CONN_RESET );
```
I would expect to see something like this in `MbedTLS.jl/ssl.jl` as well to exit the handshake flow but but I cannot find it, perhaps because I am relatively new to Julia and the underlying frameworks.

samoconnor:
Fix #164
The original version of this patch (posted here https://github.com/JuliaWeb/MbedTLS.jl/issues/164#issuecomment-422995908) resulted in "MbedTLS error code -29312: SSL - The connection indicated an EOF" when attempting to close the connection normally.
This version checks for locally initiated close.

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -111,23 +111,25 @@ end
 
 function f_send(c_ctx, c_msg, sz)
     jl_ctx = unsafe_pointer_to_objref(c_ctx)
-    return Cint(unsafe_write(jl_ctx, c_msg, sz))
+    return Cint(unsafe_write(jl_ctx.bio, c_msg, sz))
 end
 
 function f_recv(c_ctx, c_msg, sz)
     jl_ctx = unsafe_pointer_to_objref(c_ctx)
-    n = bytesavailable(jl_ctx)
+    n = bytesavailable(jl_ctx.bio)
     if n == 0
-        return Cint(MBEDTLS_ERR_SSL_WANT_READ)
+        return isopen(jl_ctx.bio) ? Cint(MBEDTLS_ERR_SSL_WANT_READ) :
+                    jl_ctx.isopen ? Cint(MBEDTLS_ERR_NET_CONN_RESET) :
+                                    Cint(n)
     end
     n = min(sz, n)
-    unsafe_read(jl_ctx, c_msg, n)
+    unsafe_read(jl_ctx.bio, c_msg, n)
     return Cint(n)
 end
 
 function set_bio!(ssl_ctx::SSLContext, jl_ctx::T) where {T<:IO}
     ssl_ctx.bio = jl_ctx
-    set_bio!(ssl_ctx, pointer_from_objref(ssl_ctx.bio), c_send[], c_recv[])
+    set_bio!(ssl_ctx, pointer_from_objref(ssl_ctx), c_send[], c_recv[])
     nothing
 end
 


src/ssl.jl
```diff
@@ -111,23 +111,25 @@ end
...
 end
 
 function f_recv(c_ctx, c_msg, sz)
     jl_ctx = unsafe_pointer_to_objref(c_ctx)
-    n = bytesavailable(jl_ctx)
+    n = bytesavailable(jl_ctx.bio)
     if n == 0
-        return Cint(MBEDTLS_ERR_SSL_WANT_READ)
+        return isopen(jl_ctx.bio) ? Cint(MBEDTLS_ERR_SSL_WANT_READ) :
+                    jl_ctx.isopen ? Cint(MBEDTLS_ERR_NET_CONN_RESET) :
```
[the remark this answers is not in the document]
samoconnor: `jl_ctx.bio` is the `TCPSocket`. If it is still open, then we want more reading to happen.
`jl_ctx.isopen == true` means that we didn't call `close(::SSLContext)` so the other end must have reset the connection.
`jl_ctx.isopen == true` means that we are expecting the connection to be closed, and it is, so there is no problem. Just return zero.

src/ssl.jl
```diff
@@ -111,23 +111,25 @@ end
 
 function f_send(c_ctx, c_msg, sz)
     jl_ctx = unsafe_pointer_to_objref(c_ctx)
-    return Cint(unsafe_write(jl_ctx, c_msg, sz))
+    return Cint(unsafe_write(jl_ctx.bio, c_msg, sz))
```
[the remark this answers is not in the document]
samoconnor: Both are needed. One writes encrypted data to the `TCPSocket`, the other writes plaintext data to the TLS library.

quinnj: I think we should add a similar `isopen` check here and return `MBEDTLS_ERR_NET_CONN_RESET` if closed; similar to the logic [here](https://github.com/ARMmbed/mbedtls/blob/8266acacc8d6e1c65fba9a048f56339d0827b2fe/library/net_sockets.c#L635)

src/ssl.jl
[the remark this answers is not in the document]
quinnj: Got it, thanks.

src/ssl.jl
samoconnor: Hmmm, interesting.
I wonder if there is a use-case where it makes any difference.
Writing to a closed stream will throw an error, so we'll never return from this function so maybe the return value doesn't matter. I guess the difference will be throwing a MbedTLS error vs a Base.IOError.
 (Hopefully soon the error will be an IOError: https://github.com/JuliaLang/julia/pull/29274/files)

samoconnor: I certainly can't see why it would hurt to return MBEDTLS_ERR_NET_CONN_RESET per the example you referred to.

quinnj: I can push that fix if you'd like; I'm pretty eager to get this fixed 😃

samoconnor: Sounds good, why don't you merge this, PR, then just push the `f_send` fix after that.

iamed2:
Is it possible to backport to a 0.6-compatible version?

PR #173, opened 2018-10-07
WIP: Replace  Base.start_reading(ctx.bio) task with Monitor Task

samoconnor:
This is an attempt to address a problem of intermittent hang-ups when doing large volumes of AWS API calls through AWSCore and HTTP.jl.

There are a few bad smells around. First, the original design of the library didn't really account for OOB messages like TLS close_notify (and other internal TLS messages that client and server may send to each other). This is why we ended up with the `@async while isopen(ctx)` loop above, and is also the reason for some convoluted code to do zero-length reads (which pump the TLS library's internal state) before doing bytes available queries. Another aspect of this that seems not quite is the fact that `eof(::SSLContext)` calls `eof(::TCPSocket)` directly. It all feels a bunch of work-arounds for an initial assumption that there would be more of a 1-1 correspondence between encrypted data arriving and decrypted data being ready to read. In face some encrypted data does not lead to decrypted data at all (OOB messages) and sometimes encrypted data has to accumulate more before it can be decrypted.

This PR removes the `@async while isopen(ctx)` loop and instead implements a `monitor` function who's job it is to watch the state of the connection, do zero-length reads as needed to pump the TLS state machine, and respond gracefully to TLS close_notify messages.

The docstring says:

>For as long as the SSLContext is open:
> - Notify readers (blocked in eof()) when decrypted data is available.
> - Check the TLS buffers for encrypted data that needs to be processed.
>   (zero-byte ssl_read(), see https://esp32.com/viewtopic.php?t=1101#p4884)
> - If the peer sends a close_notify message or closes then TCP connection,
>   then notify readers and close the SSLContext.
 >- Wait for more encrypted data to arrive.

>State management:
> - `ctx.isopen` is set `false` when `unsafe_read` or `monitor` throw an error
>    or when the `monitor` determines that the peer has closed the connection.
> - `close(::TCPSocket)` is called only at the end of the `monitor` loop.
> - `close(::SSLContext)` just calls `ssl_close_notify`.

### Summary of Changes

A `decrypted_data_ready::Condition` is added to the `SSLContext` struct so that `eof(::SSLContect)` now does `wait(ctx.decrypted_data_ready)` instead of `eof(::TCPSocket)`.

A `decrypted_data_wanted::Condition` allows readers to wake up the `monitor` in the event that it has gone to sleep because the TLS read buffer filled up while no one was reading.

`close_notify_sent` flag is added to keep track of the case where our side has asked for the connection to be closed.

A few of the `ccalls` have been pushed into wrapper functions (and the `@lockdata ctx begin` has been pushed into the wrapper functions too, to make the call-sites a bit easier to read).

When `ssl_unsafe_read` needs to wait for more data, it no longer waits on the `TCPSocket`, it waits on the `ctx.decrypted_data_ready` condition.

`Base.readavailable` no longer calls `bytesavailable` to decide how much to read. Instead it creates a buffer for MBEDTLS_SSL_MAX_CONTENT_LEN bytes (16k) and reads as much as it can into that. This removes the old behaviour of first doing a zero-length read to figure out how many bytes are available and then reading that many bytes. The revised scheme should be more efficient because it allows for reading more bytes that might arrive while the read loop is running. Most reads end up being very close to 16k once a connection settles down to a steady streaming state.

`Base.isopen` no longer looks at the `TCPSocket` open state. It only looks at the internal flags `ctx.isopen && !ctx.close_notify_sent`. Closing the `TCPSocket`, and caring if it is closed is up to the `monitor`.

`Base.close` no longer closes the `TCPSocket`, it just sends a TLS close_notify and set the `ctx.close_notify_sent` flag. It is up to the `monitor` to close the socket.

(and it doesn't get hung up anymore in my tests).
(but I am yet to try it out in more varied contexts aside from the one test case I've been focussing on, so it should be considered a WIP)

### DIFF
--- a/src/MbedTLS.jl
+++ b/src/MbedTLS.jl
@@ -13,6 +13,8 @@ end
     end
 end
 
+const DEBUG_LEVEL = 0
+
 export
 # Message digests
     MD_NONE,
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -1,3 +1,5 @@
+# Data Structures
+
 mutable struct SSLConfig
     data::Ptr{Cvoid}
     rng
@@ -26,12 +28,19 @@ mutable struct SSLContext <: IO
     datalock::ReentrantLock
     config::SSLConfig
     isopen::Bool
+    close_notify_sent::Bool
+    decrypted_data_wanted::Condition
+    decrypted_data_ready::Condition
     bio
 
     function SSLContext()
         ctx = new()
         ctx.data = Libc.malloc(1000)  # 488
         ctx.datalock = ReentrantLock()
+        ctx.isopen = false
+        ctx.close_notify_sent = false
+        ctx.decrypted_data_wanted = Condition()
+        ctx.decrypted_data_ready = Condition()
         ccall((:mbedtls_ssl_init, libmbedtls), Cvoid, (Ptr{Cvoid},), ctx.data)
         @compat finalizer(ctx->begin
             ccall((:mbedtls_ssl_free, libmbedtls), Cvoid, (Ptr{Cvoid},), ctx.data)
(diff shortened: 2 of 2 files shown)


samoconnor:
Revised PR coming soon...

PR #175, opened 2018-10-10
Resolve #174 - Step 1. Cosmetic Changes 

### LINKED ISSUE #174: AWS API calls through AWSCore.jl/HTTP.jl/MbedTLS.jl hanging
We have had an issue of intermittent hang-ups when doing large volumes of AWS API calls through AWSCore and HTTP.jl.

There are a few bad smells around. First, the original design of the library didn't really account for OOB messages like TLS close_notify (and other internal TLS messages that client and server may send to each other). This is why we ended up with the `@async while isopen(ctx)` loop above, and is also the reason for some convoluted code to do zero-length reads (which pump the TLS library's internal state) before doing bytes available queries. Another aspect of this that seems not quite is the fact that `eof(::SSLContext)` calls `eof(::TCPSocket)` directly. It all feels at this point like a bunch of work-arounds (many of them implemented by me in haste) for an initial assumption that there would be more of a 1-1 correspondence between encrypted data arriving and decrypted data being ready to read.

PRs to follow...

samoconnor:
This commit makes no semantic changes.

The order that method source code appears in the file has changed but no code within a method has been changed. These changes were made in the process of reviewing and straightening out all of the C function return code handling and state accounting code in `ssl.jl` to resolve #174.

**Changes in this PR**

Added section heading comments to ssl.jl:
 - `# Data Structures`
 - `# Handshake`
 - `# Base ::IO Connection State Methods`
 - `# Sending Data`
 - `# Sending Encrypted Data`
 - `# Receiving Data`
 - `# Receiving Encrypted Data`
 - `# Base ::IO Write Methods`
 - `# Base ::IO Read Methods`
 - `# Configuration`
 - `# C API`

Moved some methods according to section headings.

Added some docstrings.

**Next Steps**

A second PRs will follow to move `ccalls` that need `@lockdata` into wrapper functions under the `# C API` heading.

A third PR will split the implementation of the `Base.unsafe_read` and `Base.unsafe_write` functions into private `ssl_unsafe_read` and `ssl_unsafe_write`.

A final PR will introduce new state management flags and a new scheme for eof detection using the new `mbedtls_ssl_check_pending` API.

**0.6 branch vs master**

The system where #174 originates is based on Julia 0.6, which is why this PR targets that branch.
At this point there are no differences between the `0.6` branch and `master` in `ssl.jl` except for the removal of compat stuff. So, merging these changes to master later should be straightforward.

_Note, the the branch called `0.6` is a release branch for versions greater than `0.5.12`, i.e. it might have been called `release-0.5`. The name refers to the fact that the `0.5.12+` releases support Julia 0.6 whereas `0.6.0+` does not._

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -1,3 +1,5 @@
+# Data Structures
+
 mutable struct SSLConfig
     data::Ptr{Cvoid}
     rng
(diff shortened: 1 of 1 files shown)


samoconnor:
Note to reviewers, the code order changes in this PR make the github diff view completely unhelpful. However, these changes should make the diffs for following PRs easier to follow.

Probably the best way to check that this PR does not change anything other than method order is to open the [old ssl.jl](https://github.com/JuliaWeb/MbedTLS.jl/blob/73fcefe74c4730e8c321d45a435d9621f5c460c3/src/ssl.jl) and the [new ssl.jl](https://github.com/JuliaWeb/MbedTLS.jl/blob/2429a65f991723d8d617e79fe09b998e52168282/src/ssl.jl) side by side.

PR #176, opened 2018-10-10
Resolve #174 - Step 2. ccall/@lockdata segregation

### LINKED ISSUE #174: AWS API calls through AWSCore.jl/HTTP.jl/MbedTLS.jl hanging

samoconnor:
This PR make no semantic changes. It only moves some lines of code into seperate functions.

This PR moves ccalls that need @lockdata into wrapper functions under the # C API heading.

This makes the call sites easier to read and provides a place for docstrings with information on dealing with C-function return codes.

The scope of the `@lockdata` block in `Base.close` is smaller now that is has been pushed into `ssl_close_notify`, however, the calls that are no longer in the `@lockdata` block (`isopen(::TCPSocket)`, and `close(::TCPSocket)` have no interaction with the `SSLContext.data` that the lock protects.

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -43,27 +43,12 @@ mutable struct SSLContext <: IO
     end
 end
 
-macro lockdata(ctx, expr)
-    esc(quote
-        lock($ctx.datalock)
-        @assert $ctx.datalock.reentrancy_cnt == 1
-        try
-            $expr
-        finally
-            unlock($ctx.datalock)
-        end
-    end)
-end
-
 
 # Handshake
 
 function handshake(ctx::SSLContext)
     while true
-        n = @lockdata ctx begin
-            ccall((:mbedtls_ssl_handshake, libmbedtls), Cint,
-                  (Ptr{Cvoid},), ctx.data)
-        end
+        n = ssl_handshake(ctx)
         if n == 0
             break
         end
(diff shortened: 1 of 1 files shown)


samoconnor:
Ref: #174

src/ssl.jl
```diff
@@ -155,23 +140,23 @@ end
...
             catch
             end
             close(ctx.bio)
         end
         ctx.isopen = false
-    end
     nothing
 end
 
+if isdefined(Compat, :Sockets)
```
quinnj: Shouldn't be needed on 0.6? No harm leaving it in though

samoconnor: Ish...
Testing the the bowels of a deployed system with an old version of Compat it was needed.
I've found that "just upgrading" Compat in a system that is using a much older version is not always a happy experience.

PR #179, opened 2018-10-10
Resolve #174 - Step 5. Debug logging 🤖

samoconnor:
Adds debug logging with timestamp and task hash annotations (matches format produced by `const DEBUG_LEVEL = ` in HTTP.jl).
Using non-text symbols in logs is not to everyone's taste. However, the ability to ask a user to turn on this level of fine-grained logging when they have a hard-to-pin-down timing related issue has been invaluable in supporting HTTP. Also, the different colours and shapes of the symbols really helps to spot odd patterns in a stream of log data.

Below is an example with logging enabled in both HTTP.jl and here:

```
julia> using HTTP ; HTTP.get("https://httpbin.org/ip")
DEBUG: 2018-10-10T21:37:05.219 baa3 🗑  Deleted:        💀    1↑     1↓   19s www.google.com:443:61496 ≣16 RawFD(-1)
DEBUG: 2018-10-10T21:37:05.22  baa3 SSL connect: httpbin.org:443...
DEBUG: 2018-10-10T21:37:05.22  baa3 TCP connect: httpbin.org:443...
MBTLS: 2018-10-10T21:37:05.594 baa3 🤝 ...
MBTLS: 2018-10-10T21:37:05.594 baa3 f_send ➡️  404
MBTLS: 2018-10-10T21:37:05.594 baa3 f_recv WANT_READ
MBTLS: 2018-10-10T21:37:05.594 baa3 🤝  ⌛️
MBTLS: 2018-10-10T21:37:05.901 baa3 f_recv ⬅️  5
MBTLS: 2018-10-10T21:37:05.901 baa3 f_recv ⬅️  87
MBTLS: 2018-10-10T21:37:05.902 baa3 f_recv ⬅️  5
MBTLS: 2018-10-10T21:37:05.902 baa3 f_recv ⬅️  2744
MBTLS: 2018-10-10T21:37:05.903 baa3 f_recv ⬅️  5
MBTLS: 2018-10-10T21:37:05.903 baa3 f_recv ⬅️  401
MBTLS: 2018-10-10T21:37:05.903 baa3 f_recv ⬅️  5
MBTLS: 2018-10-10T21:37:05.904 baa3 f_recv ⬅️  4
MBTLS: 2018-10-10T21:37:05.921 baa3 f_send ➡️  143
MBTLS: 2018-10-10T21:37:05.922 baa3 f_send ➡️  6
MBTLS: 2018-10-10T21:37:05.922 baa3 f_send ➡️  45
[...]

PR #180, opened 2018-10-12
Resolve Issue 174 on master branch

samoconnor:
See #174, #175, #176, #178, #179


quinnj:
I'm going to do some testing off master on my end over the next day or two and then I say we tag a new release.

samoconnor:
Best to try it out with some real use before releasing. The tests are pretty lightweight in this package.
I started sketching out a test that might be able to reproduce the original issue, but it isn't done yet and there are other priorities... https://github.com/JuliaWeb/HTTP.jl/blob/master/test/streaming_combinations.jl

Issue #193, opened 2019-01-27
Undefined symbol: mbedtls_x509_crt_verify_restartable


maleadt:
I've been encountering the following on my GitLab CI while doing coverage submission on Julia 1.1:

```
julia: symbol lookup error: /builds/JuliaGPU/CUDAnative.jl/.julia/packages/MbedTLS/XkQiX/deps/usr/lib/libmbedtls.so: undefined symbol: mbedtls_x509_crt_verify_restartable
```

Full log: https://gitlab.com/JuliaGPU/CUDAnative.jl/-/jobs/151716689 -- it has worked before on 1.1 though, see eg. https://gitlab.com/JuliaGPU/CUDAnative.jl/-/jobs/149248223, but quite some packages seem to have been upgraded since (HTTP.jl, Coverage.jl, etc).

This is on a very basic Ubuntu 18.04 image with Julia binaries installed: https://github.com/JuliaGPU/gitlab-ci/blob/master/images/base/v1.1/Dockerfile

leethargo:
I have the same issue (see [travis log](https://travis-ci.org/leethargo/PipeLayout.jl/jobs/485037998)) based on Ubuntu Xenial (16.04) offered by Travis.

DilumAluthge:
I have the same error ([full log](https://api.travis-ci.com/v3/job/173244732/log.txt)) on Travis.

This is the line with the error:
```
julia: symbol lookup error: /home/travis/.julia/packages/MbedTLS/XkQiX/deps/usr/lib/libmbedtls.so: undefined symbol: mbedtls_x509_crt_verify_restartable
```

These are the lines immediately preceding:
```
Coverage.process_file: Detecting coverage for src/version.jl
Coverage.process_folder: Skipping version.jl.3853.cov, not a .jl file
Coverage.process_file: Detecting coverage for src/welcome.jl
Coverage.process_folder: Skipping welcome.jl.3853.cov, not a .jl file
Codecov.io API URL:
https://codecov.io/upload/v2?&service=travis-org&branch=master&commit=f3ec60c9a438ab1c3a9e51b750f44f288452b24f&pull_request=false&job=173244732&slug=UnofficialJuliaMirror/MirrorUpdater.jl&build=259.22
```

Version information:
- Julia Version 1.0.3
- MbedTLS.jl v0.6.7
- Coverage.jl v0.6.0

curtd:
[...]

PR #194, opened 2019-02-05
Fix dlopen order, call check_deps() in __init__

tkoolen:
Found while investigating https://github.com/JuliaWeb/MbedTLS.jl/issues/193. This still doesn't fix HTTP.jl tests, but I think these are good changes nonetheless.


visr:
The dlopen order should also be fixed when switching to a new build with https://github.com/JuliaWeb/MbedTLSBuilder/commit/5498d779237a937231b65e01b3c01e92aca3197c.

But the `check_deps` should definitely go in.

quinnj:
This is fixed on master

Issue #202, opened 2019-04-26
Support for verification X.509 certificates


sambitdash:
Expose: 

1. `mbedtls_x509_crt_verify()`
2. `mbedtls_x509_crt_verify_with_profile()`
3. `mbedtls_x509_crt_verify_info()` 

Any other method needed while implementing the above.

PR #203, opened 2019-04-29
Limited implementation of the cert_verify

sambitdash:
Partial implementation of #202.

@quinnj Please review. Few questions:
1. The callbacks for `mbedtls_x509_crt_verify` may be needed. The current implementation does not provide the intermediate steps to fix errors in case of failure. 
2. The `MbedTLS` APIs do not expose the v3 extensions properly.  We will need to expose the full ASN.1 module if we need to expose APIs for extracting the `CRLDistributionPoint` or any such APIs.


sambitdash:
Can you suggest a better way to integrate the C++ code to extract information? It seems to break in some platforms.

sambitdash:
@quinnj will you be able to help me complete this PR. I am not able to resolve the build breaks on some platforms.

sambitdash:
No plans of using this feature. So closing.

PR #204, opened 2019-06-18
Support Julia 1.2

quinnj:
Should fix https://github.com/JuliaLang/julia/issues/32220

@staticfloat, does this look right to you?


.travis.yml
```diff
@@ -1,20 +1,26 @@
...
 notifications:
   email: false
-script:
-    - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-    - julia -e 'using Pkg; Pkg.clone(pwd()); Pkg.build("MbedTLS")'
-    - julia -e 'using Pkg; Pkg.test("MbedTLS", coverage=true)'
+
 after_success:
-    - julia -e 'cd(using Pkg; Pkg.dir("MbedTLS")); Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
+   - julia -e 'ENV["TRAVIS_JULIA_VERSION"] != "1.1" && ENV["TRAVIS_OS_NAME"] != "linux" && exit(); using Pkg; Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
```
staticfloat: I don't think you need this `TRAVIS_JULIA_VERSION` thing anymore; you're only using `1.2+` above.

Project.toml
```diff
@@ -0,0 +1,20 @@
...
+uuid = "739be429-bea8-5141-9913-cc70e7f3736d"
+
+[deps]
+BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
+
+[compat]
+julia = "1.2.0"
```
staticfloat: I think you only want `julia = "1.2"` here, not `julia = "1.2.0"`.

KristofferC:
Should we upper bound the supported julia version for previous MbedTLS versions in the registry?

quinnj:
Yes, we'll need to do that, otherwise pre-1.2 julia versions will run into the same issue.

PR #210, opened 2019-11-11
Adding ca_chain_with_root_store! to ssl.jl

phyatt-corp:
Recently I had a requirement to support reading certificates out of the Windows root certificate stores ("CA", "AuthRoot", "Root", "TrustedPublisher") instead of the `cacert.pem` file for our Julia application.

I added that functionality and made it the default for Windows (see line 135 in MbedTLS.jl), so it would easily get used in HTTP.jl .  This could be made the non-default to make this have a smaller impact; also I only worked with x64 windows machines.

It respects the CRLs found in Windows, but doesn't do anything with OCSPs or OCSP Stapling.  Also I discovered a `DISALLOWED_FILETIME` property while comparing my certificates found from Windows API calls to the ones listed in the Windows Certificate Manager.  (https://github.com/ARMmbed/mbedtls/issues/2914)


peteristhegreat:
Fixes https://github.com/JuliaLang/MbedTLS.jl/issues/200 for Windows.

PR #213, opened 2020-01-10
remove duplicated info in Project

KristofferC:


quinnj:
Hmmm.....I no longer have commit/merge rights here since moving from JuliaWeb.

PR #216, opened 2020-02-24
Use MbedTLS_jll instead of BinaryProvider

JackDunnNZ:
While https://github.com/JuliaLang/julia/pull/32706 is pending, it seems like it would be good to switch this over to the JLL package in the meantime (and the stdlib change won't happen for Julia 1.3 or 1.4).

This is adapted from @quinnj's work in [`jq/stdlib`](https://github.com/JuliaLang/MbedTLS.jl/tree/jq/stdlib) but uses the JLL rather than the Julia-provided library

cc @kmsquire who asked for this here: https://discourse.julialang.org/t/packagecompiler-jl-complaining-that-mbedtls-jl-might-not-be-relocatable/34878


KristofferC:
Need to update julia compat in Project.toml and update what versions Travis tests.

JackDunnNZ:
Whoops, thanks. I also removed the FORCE_BUILD builds since that's not relevant anymore

.travis.yml
```diff
@@ -1,20 +1,13 @@
 language: julia
 
-sudo: false
-
 os:
   - linux
   - osx
```
quinnj: add `- windows` here

appveyor.yml
```diff
@@ -1,6 +1,6 @@
 environment:
```
quinnj: I'd prefer to just get rid of appveyor.yml and test 32-bit/64-bit windows via travis. It's nice to have all the OS and platforms in a single CI run.

.travis.yml
JackDunnNZ: Done

appveyor.yml
JackDunnNZ: Makes sense, removed

kmsquire:
I'll add my thanks!  Can someone please tag a release?

PR #219, opened 2020-02-28
Read default cert at compile time

### LINKED ISSUE #218: SystemError opening cert.pem with PackageCompiler-generated binary
In this line:

https://github.com/JuliaLang/MbedTLS.jl/commit/f45903118feca9aab1993a40aa82b51444c40256#diff-3f2c05e28eda33cb95585f4259519af5R758

the absolute path to `cert.pem` is recorded.

When compiling under `PackageCompiler`, that points to `<MyPackageCompiled>/share/julia/cert.pem`, which doesn't exist.

(I got around it for now by copying the `cert.pem` file to the expected location...)

Cc: @JackDunnNZ @KristofferC

JackDunnNZ:
Fixes #218 

@kmsquire can you check if this fixes the problem for you?

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -485,7 +485,7 @@ function rng!(config::SSLConfig, rng::AbstractRNG)
     rng!(config, c_rng[], rng)
 end
 
-function ca_chain!(config::SSLConfig, chain=crt_parse_file(CERT[]))
+function ca_chain!(config::SSLConfig, chain=crt_parse(DEFAULT_CERT))
     config.chain = chain
     ccall((:mbedtls_ssl_conf_ca_chain, libmbedtls), Cvoid,
         (Ptr{Cvoid}, Ptr{Cvoid}, Ptr{Cvoid}),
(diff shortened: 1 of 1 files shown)


KwatMDPhD:
Hi @JackDunnNZ , it looks like I still see this error where `cert.pem` is missing. Could you please help?

https://github.com/KwatMDPhD/GSEA.jl/issues/59

JackDunnNZ:
@KwatMDPhD The problem seems to be that PackageCompiler doesn't bundle `share/julia/cert.pem` when creating an application, whereas MbedTLS assumes this file is always present. PackageCompiler should probably be changed to bundle this file into compiled applications

KwatMDPhD:
Thanks for the explanation. @fredrikekre could you please help?

KwatMDPhD:
@JackDunnNZ , it looks like `PackageCompiler` upgraded with some breaking changes. Does `MbedTLS` take these into account these changes? For example the changes about artifacts.

KristofferC:
What version of MbedTLS is used? The point of this PR is to read the cert at compile time and not runtime so I don't see why PackageCompiler has to bundle anything.

KwatMDPhD:
Thanks for the response @KristofferC!

Screen Shot 2022-08-05 at 11 44 24

1.1.3.

JackDunnNZ:
Looks like this code has since been changed, most recently in #246, so now the cert is being read from `share/julia` at init time:

https://github.com/JuliaLang/MbedTLS.jl/blob/22f5393176a1a470317fa831c2db2392a9ccfb1e/src/ssl.jl#L803-L808

So an alternative to PackageCompiler bundling it is to go back to reading the fallback cert at compile time, as this PR originally changed

KwatMDPhD:
I want to help fix this. Can I just revert this? https://github.com/JuliaLang/MbedTLS.jl/pull/246/commits/71bd43a05575fa332e952cea49ef50f4091c479b

JackDunnNZ:
@KristofferC I was looking at how the bundled cert is used in base, and it seems that MozillaCACerts_jll assumes it is always present at init time:

https://github.com/JuliaLang/julia/blob/fa986d9e0525eedd7706929d549c61179e6c6090/stdlib/MozillaCACerts_jll/src/MozillaCACerts_jll.jl#L20

If I understand correctly, this would mean a compiled app can currently fail if it depends on MozillaCACerts_jll, since `cacert` will point to a file that doesn't exist. If that's the case, it seems like it would be a reason to bundle the cert file in PackageCompiler?

KristofferC:
Yes, that would solve it.

KwatMDPhD:
I know you are busy, but it would be amazing if you could update PackageCompiler @KristofferC 🙏

PR #227, opened 2020-10-27
These tests didn't do what you thought they did

Keno:
The `rng` assignment inside the closure was actually reassigning the local `rng`,
which would then occaisionally segfault, because `RandomDevice` is immutable,
so there's nothing that actually rooted it to the device object.

Side note: Somebody should probably add assertions that the various things
we're trying to protect by rooting them in the context object are actually mutable.
However, I've spent too much time on this already so I'll leave that to somebody else ;).

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -92,8 +92,7 @@ let
     entropy = MbedTLS.Entropy()
 
     function entropy_func(buf)
-        rng = RandomDevice()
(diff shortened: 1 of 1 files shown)


quinnj (approved):
Wowza.....thanks for tracking this down; no doubt a huge pain. I wonder if a sufficiently smart linting pass would be able to track something like this down.

PR #232, opened 2021-02-21
fix #197

### LINKED ISSUE #197: ERROR: MbedTLS.SSLContext does not support byte I/O
Can you support this? The ordinary socket could be read by `readuntil` and 'readline' funtions. 
```sh
julia> readline(c)
ERROR: MbedTLS.SSLContext does not support byte I/O
Stacktrace:
 [1] error(::Type, ::String) at ./error.jl:42
 [2] read(::MbedTLS.SSLContext, ::Type{UInt8}) at ./io.jl:182
 [3] #readuntil#290 at ./io.jl:653 [inlined]
 [4] #readuntil at ./none:0 [inlined]
 [5] #readline#284(::Bool, ::Function, ::MbedTLS.SSLContext) at ./io.jl:370
 [6] readline(::MbedTLS.SSLContext) at ./io.jl:370
 [7] top-level scope at none:0
```

gaika:
Better have an inefficient method than no method at all.

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -483,6 +483,11 @@ function Base.readavailable(ctx::SSLContext)
     return resize!(buf, n)
 end
 
+function Base.read(s::MbedTLS.SSLContext, ::Type{UInt8})
+    buf = Vector{UInt8}(undef, 1)
+    readbytes!(s, buf)
+    return buf[1]
+end
 
 
 # Configuration


src/ssl.jl
```diff
@@ -483,6 +483,11 @@ function Base.readavailable(ctx::SSLContext)
     return resize!(buf, n)
 end
 
+function Base.read(s::MbedTLS.SSLContext, ::Type{UInt8})
+    buf = Vector{UInt8}(undef, 1)
+    readbytes!(s, buf)
+    return buf[1]
```
quinnj, line 489: Hmmmm, yeah, this will be pretty inefficient. I'd rather have some kind of internal reading buffer in SSLContext that we populate and can read from.

PR #238, opened 2022-07-01
Switch default cert to MozillaCACerts_jll

quinnj:


fredrikekre:
Looks like this caused https://discourse.julialang.org/t/precompile-errors-with-pkg-update-or-pkg-add-for-some-packages/83917 ?

giordano:
For the record, the problem was due to the [Arch Linux Julia package](https://archlinux.org/packages/community/x86_64/julia/) being thoroughly broken as usual and having missing files (or not adjusting variables to point to existing files in Arch system): https://discourse.julialang.org/t/problems-installing-packages-cert-pm-no-such-file/83995/4

fredrikekre:
If bundling the cert has worked until now, perhaps worth at least falling back to that?

PR #241, opened 2022-07-09
Fall back to bundled cacert.pem.

fredrikekre:


fredrikekre:
Yea that seems easier, didn't know there was one, thought it was only the one from the JLL

quinnj:
Alternative PR proposed: https://github.com/JuliaLang/MbedTLS.jl/pull/246

PR #243, opened 2022-07-15
fix some subtle bugs with error handling and cleanup

vtjnash:
TLS streams can be a bit odd, since they mostly ignore errors on the
underlying (unencrypted) stream. Clean up some of that error handling
(dealt with in wait_for_encrypted_data, so this actually ends up not
passing through mbedtls even though we still could feed it through
there), improve flow control reliability (avoids damaging the underlying
stream), implement new `Base.closewrite` API

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -38,7 +38,6 @@ mutable struct SSLContext <: IO
     isreadable::Bool
     bytesavailable::Int
     close_notify_sent::Bool
-    async_exception::Union{Nothing,MbedException}
     bio
 
     function SSLContext()
@@ -49,7 +48,6 @@ mutable struct SSLContext <: IO
         ctx.isreadable = false
         ctx.bytesavailable = -1
         ctx.close_notify_sent = false
-        ctx.async_exception = nothing
         ccall((:mbedtls_ssl_init, libmbedtls), Cvoid, (Ptr{Cvoid},), ctx.data)
         finalizer(x->begin
             data = x.data
@@ -88,26 +86,7 @@ function handshake(ctx::SSLContext)
     ctx.bytesavailable = 0
     ctx.close_notify_sent = false
 
-    @async try
-        while ctx.isreadable
-            wait_for_decrypted_data(ctx)
-            while ctx.bytesavailable > 0
-                sleep(5)
-            end
-        end
-    catch e
-        ctx.async_exception = e
-    end
-
     nothing
-end
-
-function check_async_exception(ctx::SSLContext)
-    if ctx.async_exception !== nothing
-        e = ctx.async_exception
-        ctx.async_exception = nothing
-        throw(e)
-    end
 end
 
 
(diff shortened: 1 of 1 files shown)


quinnj:
Looks like `closewrite` is post-1.6 compatible

vtjnash:
yes, but I don't bother calling `closewrite` on the underlying stream, so it is just about making the name right for those versions with a small shim

thchr:
I bisected the HTTP.jl error from https://github.com/JuliaWeb/HTTP.jl/issues/896 to this commit.

quinnj:
Yeah, I think we better revert this change unless @vtjnash would have time to look into it sometime soon.

vtjnash:
You will run into other more frequent deadlocks now if you attempt to revert this, since it does fix some problematic assumptions about stream behaviors. The HTTP says `StackOverflowError`, so it looks like there is a problem with the handlers there

src/ssl.jl
```diff
@@ -382,29 +379,23 @@ end
...
+function f_recv(c_bio, buf, nbytes) # (Ptr{Cvoid}, Ptr{UInt8}, Csize_t)
     @assert nbytes > 0
     bio = unsafe_pointer_to_objref(c_bio)
     n = bytesavailable(bio)
-    if n == 0                                                                   ;@🤖 "f_recv $(isopen(bio) ? "WANT_READ" : "CONN_RESET")"
-        return isopen(bio) ? Cint(MBEDTLS_ERR_SSL_WANT_READ) : 
-                             Cint(MBEDTLS_ERR_NET_CONN_RESET)
+    if n == 0                                                                   ;@🤖 "f_recv $(isopen(bio) ? "WANT_READ" : "RECV_FAILED")"
+        return isreadable(bio) ? Cint(MBEDTLS_ERR_SSL_WANT_READ) :
+                                 Cint(MBEDTLS_ERR_NET_RECV_FAILED)
```
vtjnash, line 398: you might need to feed through the actual error here, or indicate to mbedtls somehow if it reached `eof` on the underlying bio

vtjnash: Does it support MBEDTLS_ERR_SSL_CONN_EOF? When I looked at the source, it wasn't clear to me how you indicated TCP FIN to mbedtls (before v1.3 of the TLS standard, this would have been an invalid termination, but was generally permitted in many clients and expected by some popular servers, so I assume mbedtls would support this)

quinnj:
The normal `HTTP.request` path doesn't have any mention of StackOverflow errors. If @vtjnash doesn't have time to investigate the issues that are arising from the changes he made, then I'll revert in the next 24 hours and tag a new release with the commits removed. I haven't seen or been shown specifically (only via vague mentions) code in HTTP.jl that is problematic, but if that can be specifically pointed out, I'm happy to dig in on the HTTP.jl side. I would obviously love to improve the robustness of the MbedTLS code, but we at least weren't having several opened issues pointing directly to the changed code prior to the changes here.

vtjnash:
Could you write a test for what you think is wrong? This PR passes all tests, and additionally passes the stricter verification for stream-implementation correctness that is now on Julia master

vtjnash:
Found it and added the missing handling + tested now https://github.com/JuliaLang/MbedTLS.jl/pull/248

PR #244, opened 2022-07-17
Adds a mutex to supposedly multithread-unsafe call

nlw0:
Bringing https://github.com/JuliaCloud/GoogleCloud.jl/pull/48 further up the stream.

MbedTLS seems to have known multi-thread safety issues (https://github.com/Mbed-TLS/mbedtls/issues/3391, https://github.com/Mbed-TLS/mbedtls/issues/3391). Putting some of the `ccall` calls might allow us to deal with it.

### DIFF
--- a/src/pk.jl
+++ b/src/pk.jl
@@ -15,6 +15,8 @@ mutable struct PKContext
     end
 end
 
+const MBEDTLSLOCK = ReentrantLock()
+
 function parse_keyfile!(ctx::PKContext, path, password="")
     @err_check ccall((:mbedtls_pk_parse_keyfile, libmbedcrypto), Cint,
         (Ptr{Cvoid}, Cstring, Cstring),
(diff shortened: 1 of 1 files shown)


nlw0:
The question is what are the `ccall` calls that would need to be in the mutex. Perhaps it's everything that takes a `c_rng` argument?

src/pk.jl
```diff
@@ -87,9 +89,11 @@ end
 
 function sign!(ctx::PKContext, hash_alg::MDKind, hash, output, rng)
     outlen_ref = Ref{Csize_t}(sizeof(output))
-    @err_check ccall((:mbedtls_pk_sign, libmbedcrypto), Cint,
-        (Ptr{Cvoid}, Cint, Ptr{UInt8}, Csize_t, Ptr{UInt8}, Ref{Csize_t}, Ptr{Cvoid}, Any),
-        ctx.data, hash_alg, hash, sizeof(hash), output, outlen_ref, c_rng[], rng)
+    @lock MBEDTLSLOCK begin
```
quinnj, line 92: I forgot that `@lock` wasn't defined until Julia 1.7; so we'll either need to add a compat like:
```julia
if !isdefined(Base, Symbol("@lock"))
macro lock(l, expr)
    quote
        temp = $(esc(l))
        lock(temp)
        try
            $(esc(expr))
        finally
            unlock(temp)
        end
    end
end
end
```
or unroll the macro yourself.

quinnj:
Ok, I fixed the compat and added the additional locks to encrypt/decrypt in https://github.com/JuliaLang/MbedTLS.jl/pull/245. (sorry for the new PR, but I couldn't push to your fork/branch here)

PR #248, opened 2022-07-29
handle normal stream EOF (TLS v1.3 feature)

vtjnash:
Implemented per mbedtls_ssl_read documentation:
https://github.com/Mbed-TLS/mbedtls/blob/53546ea099f6f53d0be653a64accd250e170337f/include/mbedtls/ssl.h#L2999

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -140,6 +140,7 @@ end
 
 True unless:
  - `close(::SSLContext)` is called, or
+ - `closewrite(::SSLContext)` is called, or
  -  the peer closed the connection.
 """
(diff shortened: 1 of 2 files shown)


thchr:
This indeed fixes the bug I had in #https://github.com/JuliaWeb/HTTP.jl/issues/896. Thanks!

(`HTTP.download("https://www.cryst.ehu.es/")` still throws a `StackOverflowError` on this PR - but that actually seems to happen on earlier versions of MbedTLS as well)

quinnj (approved):
Thank you @vtjnash; I can also confirm the fix in behavior.

PR #257, opened 2022-09-15
Use Base.IOError more consistently for read/write operation failures

quinnj:
Previously we might throw ArgumentError, MbedException, or AssertionError. This PR proposes we more closely follow Sockets stdlib behavior by making these errors Base.IOError for consistency.

Fixes https://github.com/JuliaLang/MbedTLS.jl/issues/188

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -166,23 +166,6 @@ Base.nb_available(ctx::SSLContext) = ctx.bytesavailable
 end
 
 """
-    eof(ctx::SSLContext)
-
-True if not `isreadable` and there are no more `bytesavailable` to read.
-"""
-function Base.eof(ctx::SSLContext)
-    ctx.bytesavailable > 0 && return false
-    wait_for_decrypted_data(ctx)
-    # note that the following are racy when there are multiple concurrent
-    # users of an `SSLContext`, but we're at least not going to return
-    # true until ctx.isreadable is false, which means we received a
-    # close_notify, the underlying connection was closed, or some
-    # other fatal ssl error occurred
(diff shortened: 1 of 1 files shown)


src/ssl.jl
```diff
@@ -372,8 +361,7 @@ function ssl_unsafe_read(ctx::SSLContext, buf::Ptr{UInt8}, nbytes::UInt)
                 ctx.bytesavailable = 0                                          ;@😬 "ssl_read ⌛️ $nread"
                 return nread
             elseif n < 0
-                ssl_abandon(ctx)
```
gustafsson, line 375: Was it intentional to drop the call to `ssl_abandon` here?

quinnj: Hmmmm, good question. I think the original intent was just to throw an `IOError` instead of an `MbedException`, but we perhaps want the other cleanup that happens in `ssl_abandon`, so maybe it would be better if we changed `ssl_abandon` to just throw the `IOError` instead.

gustafsson: Cleanup sounds like the right thing to do. I guess the scenario where it would make any difference would be to give a more clear error message were someone to catch the IO-error and then try to read or write again?

I wouldn't expect `ssl_abandon` to throw though. It would make it harder for me to reason about the control flow in `ssl_unsafe_read`.

A helper like `throw(mbed_ioerr(n))` could  come in handy after calling `ssl_abandon`.

gustafsson: https://github.com/JuliaLang/MbedTLS.jl/pull/260/commits/66333548833233405e606097a7392c394b34ad51

PR #260, opened 2022-10-18
Use Base.IOError in closewrite

### LINKED ISSUE #259: MbedException or IOError?
Upon closing an idle connection from the pool in HTTP.jl it [checks for any IOErrors and ignores them](https://github.com/JuliaWeb/HTTP.jl/blob/4364f2a231f783fd91e877f9dad0d4fdef537fbd/src/ConnectionPool.jl#L270).

I was going to propose in HTTP.jl that it should also ignore `MbedException` as I ran into 
```
Unhandled Task ERROR: MbedTLS error code -80: NET - Connection was reset by peer
Stacktrace:
 [1] closewrite(ctx::MbedTLS.SSLContext)
   @ MbedTLS ~/.julia/packages/MbedTLS/yELM7/src/ssl.jl:205
 [2] close(ctx::MbedTLS.SSLContext)
   @ MbedTLS ~/.julia/packages/MbedTLS/yELM7/src/ssl.jl:175
 [3] monitor_idle_connection(c::HTTP.ConnectionPool.Connection)
   @ HTTP.ConnectionPool ~/.julia/packages/HTTP/fthyG/src/ConnectionPool.jl:266
 [4] (::HTTP.ConnectionPool.var"#2#3"{HTTP.ConnectionPool.Connection})()
   @ HTTP.ConnectionPool ./task.jl:484
```

But then noticed a move away from `MbedException` into `IOError` this commit:
https://github.com/JuliaLang/MbedTLS.jl/commit/75f38e91711c749ce18a6977eb67536d04f5b173

>  Use Base.IOError more consistently for read/write operation failures
> Previously we might throw ArgumentError, MbedException, or AssertionError. This PR proposes we more closely follow Sockets stdlib behavior by making these errors Base.IOError for consistency.

Is MbedException about to be removed from MbedTLS?

I'm unfamiliar with this repo so I'm asking here instead of proposing a PR if it would it make sense to make a corresponding change to [closewrite](https://github.com/JuliaLang/MbedTLS.jl/blob/75f38e91711c749ce18a6977eb67536d04f5b173/src/ssl.jl#L205) as was done to `ssl_unsafe_write` and `ssl_unsafe_read` in the commit above?
```
- throw(MbedException(n))
+ throw(Base.IOError(strerror(n), n))
```

gustafsson:
Instead of MbedException, in line with #257

Fixes #259

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -361,6 +361,7 @@ function ssl_unsafe_read(ctx::SSLContext, buf::Ptr{UInt8}, nbytes::UInt)
                 ctx.bytesavailable = 0                                          ;@😬 "ssl_read ⌛️ $nread"
                 return nread
             elseif n < 0
+                ssl_abandon(ctx)
                 mbed_ioerr(n)
             end
 
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -78,7 +78,7 @@ function handshake(ctx::SSLContext)
             end
         else
             ssl_abandon(ctx)                                                    ;@💀 "🤝  💥"
-            throw(Base.IOError(strerror(n), n))
+            mbed_ioerr(n)
         end
     end
                                                                                 ;@😬 "🤝  ✅"
@@ -108,7 +108,7 @@ function ssl_abandon(ctx::SSLContext)                                           
     ctx.close_notify_sent = true
     close(ctx.bio)
     n = ssl_session_reset(ctx)
-    n == 0 || throw(Base.IOError(strerror(n), n))
+    n == 0 || mbed_ioerr(n)
     nothing
 end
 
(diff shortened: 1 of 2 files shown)


gustafsson:
I stumbled upon a related question when making this change:
https://github.com/JuliaLang/MbedTLS.jl/pull/257/files#r997901371

src/ssl.jl
```diff
@@ -202,7 +202,7 @@ function closewrite(ctx::SSLContext)
                       "never returns ...WANT_READ/WRITE."
     elseif n != 0
         ssl_abandon(ctx)
```
quinnj, line 204: Yeah, as I commented on in the related PR, the problem here is `ssl_abandon` is going to throw before we're able to throw the `IOError` on the next line.

gustafsson: Thanks that make sense. However `ssl_abandon` doesn't throw in my scenario (#259) where I reliably end up here  with `n = -80, strerror(n) = NET - Connection was reset by peer`.

So the issue is resolved for me.

Would you like me to replace `MbedException(n)` with `Base.IOError(strerror(n), n)` in `ssl_abandon` anyways?

Should we also make the swap in `handshake` while at it? That's the last remaining reference to `MbedException` in this file.

quinnj: Yeah, let's switch them all to IOError; thanks.

quinnj (approved):
Thanks!

PR #265, opened 2023-10-26
Add GC-safe regions around `ssl_read` and `ssl_write`

kpamnany:
We've observed `mbedtls_ssl_read` and `mbedtls_ssl_write` show up in CPU profiles while also seeing long GC time-to-safepoint. These `ccall`s call back into Julia (`f_send` and `f_recv`), during which GC should be able to run. But I can't tell if, for large enough reads or writes, there could be a long interval without reaching a GC safepoint.

Adding GC-safe regions around the `ccall`s will let GC run even if a thread is busy inside. Since `c_send` and `c_recv` are `@cfunction`s, they will do the right thing wrt GC safety.

Ref: https://github.com/JuliaLang/julia/issues/51574

This is a temporary fix; when https://github.com/JuliaLang/julia/pull/49933 lands, this should be changed to do whatever that requires.

If a reviewer knows for certain that a thread won't remain inside the `ccall` without calling back into Julia for multiple seconds, then this PR can be discarded.

### DIFF
--- a/src/ssl.jl
+++ b/src/ssl.jl
@@ -701,11 +701,23 @@ closed.
 https://tls.mbed.org/api/ssl_8h.html#aa2c29eeb1deaf5ad9f01a7515006ede5
 """
 function ssl_read(ctx::SSLContext, ptr, n)::Int
+    ret = 0
     @lockdata ctx begin
-        return ccall((:mbedtls_ssl_read, libmbedtls), Cint,
-                     (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t),
-                     ctx.data, ptr, n)
+        # We want to allow GC to run while this thread is in the `ccall`.
+        # When https://github.com/JuliaLang/julia/pull/49933 is completed
+        # and lands, this should be changed to what is required by that.
+        ccd = Base.cconvert(Ptr{Cvoid}, ctx.data)
+        cptr = Base.cconvert(Ptr{Cvoid}, ptr)
+        GC.@preserve ccd cptr begin
+            ucd = Base.unsafe_convert(Ptr{Cvoid}, ccd)::Ptr{Cvoid}
+            ucptr = Base.unsafe_convert(Ptr{Cvoid}, cptr)::Ptr{Cvoid}
+            gc_state = @ccall(jl_gc_safe_enter()::Int8)
+            ret = ccall((:mbedtls_ssl_read, libmbedtls), Cint,
+                        (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t), ucd, ucptr, n)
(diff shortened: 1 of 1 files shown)


src/ssl.jl
```diff
@@ -701,11 +701,23 @@ closed.
...
-                     (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t),
-                     ctx.data, ptr, n)
+        ccd = Base.cconvert(Ptr{Cvoid}, ctx.data)
+        cptr = Base.cconvert(Ptr{Cvoid}, ptr)
+        GC.@preserve ccd cptr begin
+            # Allow Julia to GC while reading/decrypting
+            gc_state = @ccall(jl_gc_safe_enter()::Int8)
+            ret = ccall((:mbedtls_ssl_read, libmbedtls), Cint,
+                        (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t),
+                        Base.unsafe_convert(Ptr{Cvoid}, ccd)::Ptr{Cvoid},
```
vtjnash, line 713: Might be worth moving the unsafe_convert call outside too, just to minimize the code that is inside the unsafe region

Otherwise, seems like a decent workaround for now. Perhaps leave a comment here and on Valentin's PR that fixes this in `ccall` to revisit it?

kpamnany: I'd discussed the correct pattern for this with @vchuravy (he opened https://github.com/JuliaLang/julia/pull/51884 during that discussion) and am following what the new documentation suggests. If the `unsafe_convert`s should also be outside the `GC.@preserve` block, then maybe the documentation should be updated again?

Will add notes as suggested.

vtjnash: They probably should be outside, but it doesn't really matter if the `@GC.preserve` region starts too soon, as that would have been extended anyways in the roots-coloring pass (llvm-late-gc-lowering). It matters more here because you are calling `jl_gc_safe_enter`, which is not entirely legal to appear in Julia code, so you should be very careful what might end up inside there.

kpamnany: Like so?

vtjnash (approved):
LGTM

kpamnany:
The CI failures for Julia nightly are:
```
ERROR: LoadError: conversion to pointer not defined for Base.CodeUnits{UInt8, String}
```
They're coming from `crypt!` in `cipher.jl` [here](https://github.com/JuliaLang/MbedTLS.jl/blob/e84f9732360700a0f5baf59a65d5211411cb97fd/src/cipher.jl#L265), which I haven't touched in this PR.

vtjnash:
There are so many random `pointer` calls here that only serve to ensure incorrectness and serve no other value

PR #267, opened 2023-11-03
remove unsafe pointer calls

vtjnash:
Almost entirely unneeded, and entirely unsafe, except for the failure to specify the ccall API correctly.

### DIFF
--- a/src/error.jl
+++ b/src/error.jl
@@ -18,11 +18,12 @@ mbed_err(ret) = throw(MbedException(ret))
 mbed_ioerr(ret) = throw(Base.IOError(strerror(ret), ret))
 
 function strerror(ret, bufsize=1000)
-    buf = Vector{UInt8}(undef, bufsize)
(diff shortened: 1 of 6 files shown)


src/error.jl
```diff
@@ -18,11 +18,12 @@ mbed_err(ret) = throw(MbedException(ret))
 mbed_ioerr(ret) = throw(Base.IOError(strerror(ret), ret))
 
 function strerror(ret, bufsize=1000)
-    buf = Vector{UInt8}(undef, bufsize)
+    buf = Base.StringVector(bufsize)
     ccall((:mbedtls_strerror, libmbedcrypto), Cint,
```
yuyichao, line 22: This is completely random but it seems that the return type is void….

vtjnash:
```suggestion
    ccall((:mbedtls_strerror, libmbedcrypto), Cvoid,
```

nsslh:
Regression: #274.

PR #273, opened 2023-11-09
Update TagBot.yml

vtjnash:
When at first you don't succeed, try harder?


kpamnany (commented):
Those lines came directly from [here](https://github.com/JuliaRegistries/TagBot/blob/master/example.yml) which should be right?

vtjnash:
I am not sure, since it is still failing to run

